PR #4, opened 2020-01-13
Start CI migration to Jenkins pipelines

ArsenArsen:
Possible defect: the two rsync steps towards the end could make the builds fail on computers of developers not set up right. This is a problem only for those who wish to maintain CI

### DIFF
--- a/ci/Jenkinsfile
+++ b/ci/Jenkinsfile
@@ -1,0 +1,48 @@
+node {
+    cleanWs()
+    dir('src') { checkout scm }
+    def userId = sh(returnStdout: true, script: 'id -u').trim()
+    docker.build('managarm_buildenv', "--build-arg USER=${userId} src/docker/")
+          .inside {
+        dir('build') {
+            stage('Build system') {
+                sh '''#!/bin/sh
+                set -xe
+                xbstrap init ../src || true
+                xbstrap install --all
+                '''
+            }
+            stage('Archive packages') {
+                sh 'xbstrap archive --all'
+            }
+        }
+        stage('Make docs') {
+            dir('docs') {
+                sh '''#!/bin/sh
+                set -xe
+                mkdir -p hel
+                sed 's|@ROOTDIR@|../src/managarm/hel|' ../src/managarm/hel/Doxyfile.in > Doxyfile
+                doxygen
+                '''
+            }
+        }
+    }
+
+    stage('Make image') {
+        dir('build') {
+            sh '''#!/bin/sh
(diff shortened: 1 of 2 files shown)


ci/Jenkinsfile
```diff
@@ -0,0 +1,48 @@
+node {
+    cleanWs()
+    dir('src') { git 'https://github.com/managarm/bootstrap-managarm.git' }
+    docker.build('managarm_cienv', 'src/ci/docker/').inside {
+        dir('build') {
+            stage('Build system') {
+                sh '''#!/bin/sh
+                set -xe
+                xbstrap init ../src || true
```
ArsenArsen: quick remark about this line: `|| true` is unnecessary if `cleanWs()` is used. I used it to speed up prototyping. It's not mandatory, but it should have little to no effect as the only effect `init` has is a symlink

ci/Jenkinsfile
```diff
@@ -0,0 +1,51 @@
+node {
+    cleanWs()
+    dir('src') { checkout scm }
+    def userId = sh(returnStdout: true, script: 'id -u').trim()
+    echo "${userId}"
```
avdgrinten: Left over from debugging?

docker/Dockerfile
```diff
@@ -1,16 +1,18 @@
...
 
-RUN useradd -ms /bin/bash managarm_buildenv
+RUN useradd -ms /bin/bash managarm_buildenv -u $USER
 USER managarm_buildenv
 WORKDIR /home/managarm_buildenv
 
 RUN git config --global user.email "managarm_buildenv@localhost"
 RUN git config --global user.name "managarm_buildenv"
 
+ADD gitconfig /etc/gitconfig
```
avdgrinten: Why do we still need this?

ci/Jenkinsfile
```diff
@@ -0,0 +1,51 @@
...
+                mkdir -p hel
+                sed 's|@ROOTDIR@|../src/managarm/hel|' ../src/managarm/hel/Doxyfile.in > Doxyfile
+                doxygen
+                '''
+            }
+        }
+    }
+
+    stage('Make image') {
+        withEnv(['LIBGUESTFS_BACKEND_SETTINGS=force_tcg']) {
```
avdgrinten: We don't need `force_tcg` anymore.

docker/Dockerfile
ArsenArsen: It can probably be removed without losing anything

PR #6, opened 2020-02-17
Make cairo compilable on systems with X11 headers present

no92:
Cairo compilation fails on my system, complaining about double redefinitions in my X11 system headers. Passing `--disable-xlib` fixes the issue, makes cairo compile, and testing weston does not reveal any glaring issues with this.


avdgrinten:
Good catch; thank you for the PR. LGTM.

Issue #7, opened 2020-02-19
lua: Install to /usr


avdgrinten:
Currently, lua is installed to the root directory (i.e., `/bin`). Like other general purpose programs, it should be in `/usr/bin` instead. @krkk could you look into that?

PR #8, opened 2020-02-19
lua: Install to /usr

krkk:
Closes: #7


avdgrinten:
Thank you! (Compiling Lua seems to be quite archaic with those header files and makefile options that need to be set manually...)

PR #12, opened 2020-04-11
runit: add sys/select.h patch

ArsenArsen:
this patch is to be included until upstream picks up on it in the mailing list


avdgrinten:
Thank you for the PR. Let's see if this patch is merged upstream.

PR #13, opened 2020-04-11
Added a grep port, works well, however OOM's on large recursive direc…

Dennisbonke:
…tories


avdgrinten:
LGTM. Thank you very much for the PR!

PR #19, opened 2020-04-24
Added a working port of gawk and a port of xz-utils.

Dennisbonke:
The `gawk` port works fine in my limited testing, but the `xz-utils` port needs some work in mlibc, the current issue for `xz-utils` is an unknown `printf()` terminator, see the issue [here](https://github.com/managarm/mlibc/issues/57) for more information on that.


avdgrinten:
Looks good, thanks a lot for your efforts.

PR #21, opened 2020-04-25
Add pcre and glib ports

no92:
PCRE should work, as by @Dennisbonke quick grep test.

IIRC, glib needs some outstanding mlibc changes, which will be supplied soon™

### DIFF
--- a/patches/glib/0001-Small-fixes.patch
+++ b/patches/glib/0001-Small-fixes.patch
@@ -1,0 +1,160 @@
+From e3257b91e4920fbf4753452aa05d5b9941024c8e Mon Sep 17 00:00:00 2001
+From: no92 <no92.mail@gmail.com>
+Date: Sun, 26 Apr 2020 01:05:06 +0200
+Subject: [PATCH] Small fixes
+
+---
+ gio/gnetworking.h.in       |  2 ++
+ gio/gsocket.c              | 33 +++++++++++++++++++++++++++++++++
+ gio/gthreadedresolver.c    |  7 +++++--
+ gio/meson.build            |  4 ++--
+ gio/xdgmime/xdgmimecache.c |  2 +-
+ glib/glib-init.c           |  1 +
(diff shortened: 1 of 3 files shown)


patches/glib/0001-Small-fixes.patch
```diff
@@ -0,0 +1,160 @@
...
++#endif
++#ifndef SO_NREAD
++#define SO_NREAD 0
++#endif
++#ifndef CMSG_DATA
++#define CMSG_DATA(x) ((unsigned char*) x)
++#define CMSG_FIRSTHDR(x) ((struct cmsghdr *) NULL)
++#define CMSG_NXTHDR(x, y) ((struct cmsghdr *) NULL)
++#endif
++#endif
```
avdgrinten, line 69: These can probably be (mostly?) fixed in mlibc. Maybe that's something @Dennisbonke can integrate into his mlibc PR?

patches/glib/0001-Small-fixes.patch
```diff
@@ -0,0 +1,160 @@
...
+  * and bionic/libc/private/arpa_nameser.h */
+ typedef struct {
+@@ -924,7 +924,10 @@ free_records (GList *records)
+ }
+ 
+ #if defined(G_OS_UNIX)
+-#ifdef __BIONIC__
++#if defined(__BIONIC__) || defined(__managarm__)
++#if defined(__managarm__)
++static int h_errno;
```
avdgrinten, line 94: What's `h_errno`?

no92: Similar to errno, but for `gethostbyname()` and `gethostbyaddr()`. They are marked as obsolete, so mlibc implementation might not even be needed.

For reference, see the [man page on h_errno](https://linux.die.net/man/3/h_errno).

avdgrinten:
Closed in favor of #23 (which is based on this patch). Thanks a lot for this work!

PR #24, opened 2020-05-04
Fix for python's build system and new ports

Dennisbonke:
This PR fixes python's build system when python 3.8 is not available on the host and adds the following ports.

- `desktop-file-utils`, a runtime dependency of `glib`
- `which`

### DIFF
--- a/bootstrap.yml
+++ b/bootstrap.yml
@@ -95,6 +95,18 @@ sources:
         environ:
           'NOCONFIGURE': 'yes'
 
+  - name: 'python'
+    subdir: 'ports'
+    git: 'https://github.com/python/cpython.git'
+    tag: 'v3.8.2'
+    tools_required:
+      - host-autoconf-v2.69
+      - host-automake-v1.15
+      - host-libtool
+      - host-pkg-config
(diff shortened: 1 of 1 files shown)


Dennisbonke:
This PR should also fix the build system of `file` on hosts which don't have `file` 5.38

bootstrap.yml
```diff
@@ -581,7 +626,6 @@ packages:
         - host-pkg-config
       regenerate:
         - args: ['./autogen.sh']
-          environ:
             'NOCONFIGURE': 'yes'
```
avdgrinten, line 629: This seems to be an oversight? (Without `environ`, the `NOCONFIGURE` line does nothing.)

bootstrap.yml
```diff
@@ -1498,19 +1542,9 @@ packages:
         quiet: true
 
   - name: python
-    source:
```
avdgrinten, line 1501: Hm, don't we want `from_source:` instead? (If it seems to work without that, I'd consider that a bug in xbstrap and not a feature :smiley:)

PR #27, opened 2020-05-07
Added some ports

Dennisbonke:
This PR adds the following ports

- `gmp`
- `mpfr`
- `mpc`

And fixes the following port
- `gcc`

### DIFF
--- a/bootstrap.yml
+++ b/bootstrap.yml
@@ -32,9 +32,6 @@ sources:
       - host-autoconf-v2.69
       - host-automake-v1.11
     regenerate:
-      # download_prerequisites should probably move to some "post_checkout" step.
-      - args: ['./contrib/download_prerequisites']
-        workdir: '@THIS_SOURCE_DIR@'
       - args: ['autoconf']
         workdir: '@THIS_SOURCE_DIR@/gcc'
       - args: ['autoconf']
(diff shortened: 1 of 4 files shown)


bootstrap.yml
```diff
@@ -40,6 +40,19 @@ sources:
...
+    git: 'git://gcc.gnu.org/git/gcc.git'
+    tag: 'releases/gcc-9.2.0'
+    tools_required:
+      - host-autoconf-v2.69
+      - host-automake-v1.11
+    regenerate:
+      - args: ['autoconf']
+        workdir: '@THIS_SOURCE_DIR@/gcc'
+      - args: ['autoconf']
+        workdir: '@THIS_SOURCE_DIR@/libstdc++-v3'
```
avdgrinten, line 54: Why do we need a second source for GCC?

patches/gcc/0002-Managarm-target.patch
```diff
@@ -87,6 +87,10 @@ index ddd3b8f4d..9701267b9 100644
 +    tmake_file="${tmake_file} t-slibgcc"
 +    thread_file='posix'
 +    ;;
++  *-managarm-pc*)
++    tmake_file="${tmake_file} t-slibgcc"
++    thread_file='posix'
++    ;;
```
avdgrinten, line 93: Why do we need `managarm-pc`?

avdgrinten: (Not saying that it's wrong, I'm just surprised that another suffix is needed.)

Dennisbonke: Gcc defaults to this suffix with the current configuration options. We can probably get it to accept managarm-system there as well but something-pc is pretty universal as a suffix

avdgrinten: OK.

avdgrinten:
Thank you for the PR!

PR #29, opened 2020-05-12
Add support files and port perl

Dennisbonke:
This PR adds an `/etc/passwd` and an `/etc/group` file, which bash uses to populate the prompt with an username

### DIFF
--- a/bootstrap.yml
+++ b/bootstrap.yml
@@ -1,3 +1,6 @@
+imports:
+  - file: ../src/bootstrap.d/dev-perl.yml
+  - file: ../src/bootstrap.d/sys-libs.yml
 
 repository:
   url: 'https://pkgs.managarm.org/nightly'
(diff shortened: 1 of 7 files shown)


bootstrap.yml
```diff
@@ -578,6 +578,8 @@ packages:
       - args: ['cp', '@SOURCE_ROOT@/extrafiles/.bashrc', '@THIS_COLLECT_DIR@/root']
       - args: ['cp', '@SOURCE_ROOT@/extrafiles/profile', '@THIS_COLLECT_DIR@/etc']
       - args: ['cp', '@SOURCE_ROOT@/extrafiles/bash.bashrc', '@THIS_COLLECT_DIR@/etc']
+      - args: ['cp', '@SOURCE_ROOT@/extrafiles/passwd', '@THIS_COLLECT_DIR@/etc']
+      - args: ['cp', '@SOURCE_ROOT@/extrafiles/group', '@THIS_COLLECT_DIR@/etc']
```
avdgrinten, line 582: Hm, this should probably be in its own package (`base-configuration`?) and not in the Bash package.

extrafiles/passwd
```diff
@@ -0,0 +1,3 @@
+root:x:0:0:root:/root:/bin/bash
+nobody:x:99:99:nobody:/dev/null:/usr/bin/false
```
avdgrinten, line 2: Why is nobody 99 instead of 65534?

avdgrinten:
LGTM but this branch has a conflict.

PR #37, opened 2020-06-07
Made /bin and friends symlinks to their /usr counterparts

Dennisbonke:
This PR makes `/bin`, `/lib` and `/sbin` symlinks to their counterparts in `/usr`
After merging of this PR, existing build environments should do the following to be sure:

- Either mount the image, and remove the `/bin` and `/lib` folders, or remove the image completely and generate a new one,
- Remove the `/bin` and `/lib` folders from the `system-root` directory,
- Run the `prepare-sysroot` script,
- Run `xbstrap install --all` to get the symlinks and needed configuration files,
- Update the image as usual.

After performing the above steps once, they shouldn't be needed any longer, and regular image updating can be resumed.

This PR is blocked on a new xbstrap release ~~which includes managarm/xbstrap#22 and managarm/xbstrap#23~~(both merged)

### DIFF
--- a/bootstrap.yml
+++ b/bootstrap.yml
@@ -494,8 +494,14 @@ packages:
     configure: []
     build:
       - args: ['mkdir', '-p', '@THIS_COLLECT_DIR@/etc/']
+      - args: ['mkdir', '-p', '@THIS_COLLECT_DIR@/usr/bin']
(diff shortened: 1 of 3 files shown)


avdgrinten:
LGTM and can be merged now thanks to the xbstrap release.

PR #53, opened 2020-08-30
Add version info to nearly all packages

Dennisbonke:
This PR adds versioning information for use in packaging where it was known, e.g. everything that is not built directly from master (with a few small exceptions).

### DIFF
--- a/bootstrap.d/app-misc.yml
+++ b/bootstrap.d/app-misc.yml
@@ -4,6 +4,8 @@ packages:
       subdir: ports
       git: 'https://github.com/mtoyoda/sl.git'
       branch: 'master'
+      # While we build directly from master, the Makefile suggests version 5.03.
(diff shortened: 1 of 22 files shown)


bootstrap.d/app-misc.yml
```diff
@@ -4,6 +4,8 @@ packages:
       subdir: ports
       git: 'https://github.com/mtoyoda/sl.git'
       branch: 'master'
+      # While we build directly from master, the Makefile suggests version 5.03.
+      version: '5.03'
```
avdgrinten, line 8: We should probably fix a commit in this case, using `commit:`.

avdgrinten (approved):
LGTM except for that one comment.

avdgrinten:
Nice, thank you!

PR #56, opened 2020-09-25
Add several meta packages

Dennisbonke:
This PR adds a `base` meta package which can be used to boot into kmscon and download more packages with `xbps` pending mlibc support (issues managarm/mlibc#51, managarm/mlibc#168, managarm/mlibc#169 and managarm/mlibc#170).
It also adds a `base-devel` meta package which holds the equivalent of `build-essentials` on debian, e.g essential build tools like gcc, binutils and make. As we are still missing software in the ports collection to fill this properly, I expect this meta package to grow in the future

The following ports are enabled by default:

- `xz-utils`
- `wget`
- `make`
- `m4`
- `patch`
- `pkg-config`

Furthermore, the following port received an update
- `python`, bump revision and depend on `xz-utils`

### DIFF
--- a/bootstrap.d/app-arch.yml
+++ b/bootstrap.d/app-arch.yml
@@ -104,7 +104,6 @@ packages:
           DESTDIR: '@THIS_COLLECT_DIR@'
 
   - name: xz-utils
-    default: false
     source:
       subdir: ports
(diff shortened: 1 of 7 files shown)


avdgrinten (approved):
LGTM.

PR #57, opened 2020-10-03
Update several packages

Dennisbonke:
This PR updates the following packages:

- `glib`, from 2.59.2 to 2.66.1,
- `grep`, from 3.4 to 3.5,
- `libpng`, from 1.6.34 to 1.6.37,
- `libxkbcommon`, from 0.8.0 to 1.0.1, and modified the build instructions to also build `libxkbcommon-x11.so`,
- `pixman`, from 0.34.0 to 0.40.0,
- `wayland-protocols`, from 1.18 to 1.20.


avdgrinten:
LGTM. Why can we drop these two patches? Did upstream packages fix their build systems accordingly?

avdgrinten:
As a side note, it would be cleaner to have one commit per package/tool.

Dennisbonke:
Regarding the patches, I removed the patches locally before trying to upgrade `libxkbcommon` a few weeks ago. It build fine without modifications and I've been running this version ever since.

PR #60, opened 2020-10-11
bootstrap.yml: Add stability_level annotations

Dennisbonke:
This PR prepares the repo for the new xbstrap release. It adds `everything_by_default: false` to only build a minimal subset by default (when using `xbstrap`, `xbstrap-pipeline` ignores this), it adds `default: true` to that minimal subset (which can boot `kmscon` with a `bash` shell and some config files, `core-files` for filesystem layout and `iana-etc` for the `/etc/services` and `/etc/protocols` file) and it dropped all the `default: false`, as that is now the default, and added `stability_level: broken` to packages which had `default: false`. I expect to remove this label from some packages in the coming days after I've confirmed that they actually build and do something remotely useful, like not crashing the `posix-subsystem` server on executing their help command.

This PR is blocked on a new xbstrap release (duh), but should be merged immediately after releasing, to not break nightly.

### DIFF
--- a/bootstrap.d/app-admin.yml
+++ b/bootstrap.d/app-admin.yml
@@ -1,6 +1,6 @@
 packages:
   - name: sysklogd
-    default: false
+    stability_level: broken
     source:
       subdir: 'ports'
       git: 'https://github.com/troglobit/sysklogd.git'
--- a/bootstrap.d/app-shells.yml
+++ b/bootstrap.d/app-shells.yml
@@ -1,5 +1,6 @@
 packages:
   - name: bash
+    default: true
     source:
       subdir: 'ports'
       git: 'https://git.savannah.gnu.org/git/bash.git'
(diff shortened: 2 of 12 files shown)


Dennisbonke:
As of commit `de2b2e8` packages that build and not take down the system are promoted to stable, which means this PR is waiting on the new `xbstrap` release

avdgrinten:
Hm, isn't it enough to mark only the "top-level" packages as `default: true`? xbstrap should automatically build dependencies, right?

Dennisbonke:
I haven't tested that, as I don't have a development version of xbstrap installed, but I guess that would work too. This way is more verbose for people looking at the files

avdgrinten:
It's also annoying to keep track of which package causes the dependencies. Can you update the PR to only annotate top-level packages?

avdgrinten:
LGTM now, thanks :)

PR #62, opened 2020-11-08
Fix for ci image generation and some updates

Dennisbonke:
This PR fixes the image generation for the ci and updates the `base-devel` meta package with `bc` and `cmake` now that those are available. Furthermore, a new meta package called `gui` is added, which depends on `base`, `weston` and some X based utility programs.

### DIFF
--- a/bootstrap.yml
+++ b/bootstrap.yml
@@ -1301,7 +1301,7 @@ packages:
 
       
 tasks:
-  - name: make-useful-image
+  - name: useful-image-dependencies
     pkgs_required:
       - base
       - gui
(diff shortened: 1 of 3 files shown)


bootstrap.yml
```diff
@@ -1301,10 +1301,21 @@ packages:
 
       
 tasks:
+  - name: make-useful-image
```
avdgrinten, line 1304: Maybe we should call this `useful-image-dependencies` since it doesn't actually "make" anything.

avdgrinten:
LGTM, thank you for the PR.

PR #63, opened 2020-11-14
Update readme for protobuf and mdbook

dchapiesky:


README.md
```diff
@@ -64,10 +64,14 @@ Now proceed to the Building paragraph.
...
+    `build-essential`, `pkg-config`, `autopoint`, `bison`, `curl`, `flex`, `gettext`, `git`, `gperf`, `help2man`, `m4`, `mercurial`, `ninja-build`, `python3-mako`, `python3-yaml`, `texinfo`, `unzip`, `wget`, `xsltproc`, `xz-utils`, `libexpat1-dev`, `rsync`, `python3-pip`, `python3-libxml2`, `netpbm`, `itstool`, `zlib1g-dev`, `libgmp-dev`, `libmpfr-dev`, `libmpc-dev`, `subversion`, `gawk`, `libwayland-bin`, `libpng-dev`, `gtk-doc-tools`, `groff`, `libglib2.0-dev-bin`, `ragel`.
 1.  `meson` is required. There is a Debian package, but as of Debian Stretch, a newer version is required.
-    Install it from pip: `pip3 install meson`.
+    Install it from pip: `pip3 install meson`
+1.  `protobuf` is also required. There is a Debian package, but a newer version is required.
+    Install it from pip: `pip3 install protobuf`
```
ArsenArsen, line 71: prefer a virtualenv or a user-specific install, especially when in collision with a system package (this applies for meson and the rest of the pip invocations here too)

dchapiesky: It was my understanding that pip, upon collision with a system wide version, will install in the user's home directory under .local...  it does this automatically

dchapiesky: The user simply must place ./local/bin in their path

ArsenArsen: that doesn't change much, since anything would rely on tools being in the proper location in path. it is generally safer and cleaner to use a virtualenv, anyway, on any distribution, for non-system-level package installs

dchapiesky: The comments tell the end user *what* to install... how they do it and whether or not they use virtualenv is their concern?  The update in the PR addresses the very real issue that an older version of protobuf will cause the managarm build to fail...

avdgrinten: I think the change is reasonable. How would a virtualenv help here? We already ask the user to install meson, adding instruction to enter a virtualenv would be quite verbose. Plus, if people are concerned about polluting their environment, they can just use Docker.

ArsenArsen: the comment is a more general one, not one specific to this change, this change just happened to provoke me to write it.  
and yes, I agree, if a consistent environment is a primary concern, it's smarter to use docker, however, that could lead to integrations with development tools (such as ccls and clangd, relying on compilation databases) being harder, so I can also understand why someone would want to run the tools in their native environment

> The comments tell the end user *what* to install... how they do it and whether or not they use virtualenv is their concern?

fair enough, I'd add a note to advise people to do it, though.

dchapiesky: Ah I was a little confused Arsen... sorry...  When I wrote the first draft I actually did go verbose and then trimmed it down to match the existing instructions....    for example... I had never used cargo before (for mdbook) and had to add .cargo/bin to my path.... or on another occasion having "." first in path breaks the cross-compile of make in managarm...  there is a laundry list of tweeks that are good to know about but each developer has to resolve them for their own environments.

avdgrinten:
LGTM. I will squash the commits to get rid of the merges.

PR #65, opened 2020-11-28
Various new ports

Dennisbonke:
This PR adds the following ports:

- `libtasn1`, required dependency for `p11-kit` and used to manage ASN.1 data structures,
- `p11-kit`, an interface for Public-Key Cryptography Standard 11, also used in generating the ca-certificate store,
- `xprop`, a property displayer for X,
- `xset`, a tool for setting miscellaneous X server parameters.

Furthermore, the version number for `libuv` was fixed and the list of dependencies in the readme and Dockerfile have been updated.

This PR is blocked on managarm/mlibc#187


avdgrinten:
Why do we need additional packages on the host?

Are some of the patches taken from other OSes? Do they need attribution?

Dennisbonke:
`p11-kit` generates some header files at build time using programs from `libtasn1-bin`, however due to those files going in the pkg-build dir and p11-kit looking from the source dir this fails, which is why I included the copy generated on my host as a patch. Regarding patches, the distribution specific hook was taken from Beyond Linux From Scratch, the other 2 are local work

avdgrinten:
What's the license of BLFS?

Dennisbonke:
The licenses of BLFS are Creative Commons Attribution-NonCommercial-ShareAlike 2.0 and the MIT license. I’m guessing the patch would fall under the MIT license here

avdgrinten:
But in that case the patch needs attribution. State the origin, copyright holder and license in the patches commit message.

Dennisbonke:
Yeah I just realized that. Good catch, I’ll modify that later today

avdgrinten:
Alright, LGTM.

PR #67, opened 2020-12-07
libiconv port and various updates

Dennisbonke:
This PR adds the following new ports:

- `libiconv`, as mlibc doesn't properly implement the `iconv()` functions we use this utility library to provide those.
- `man-pages-posix`, a collection of man pages in the sections `0p`, `1p` and `3p`, mostly related to posix functions.

Furthermore, the following ports were updated:

- `libtool`, updated version from 2.4.5 to 2.4.6.
- `man-db`, add a dependency on `libiconv`, bump revision and fix the install location to use the correct version.
- `libarchive`, bump revision and add a dependency on `libressl`, `zlib`, `xz-utils`, `libiconv`, `libexpat` and `libxml`
- `vim`, bump revision and add a dependency on `libiconv`.
- `glib`, bump revision and add a dependency on `zlib`.
- `fontconfig`, bump revision and add a dependency on `libiconv`.
- `SDL2`, bump revision and add a dependency on `libiconv`.

### DIFF
--- a/bootstrap.d/app-arch.yml
+++ b/bootstrap.d/app-arch.yml
@@ -57,9 +57,17 @@ packages:
       git: 'https://github.com/libarchive/libarchive.git'
       tag: 'v3.4.3'
       version: '3.4.3'
+      rev: '2'
     tools_required:
       - host-cmake
       - system-gcc
+    pkgs_required:
+      - libressl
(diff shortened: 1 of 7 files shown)


avdgrinten (approved):
LGTM.

PR #68, opened 2020-12-18
Various updates

Dennisbonke:
This PR updates the following ports:

- `nano`, update from version 5.2 to 5.4,
- `libarchive`, update from version 3.4.3 to 3.5.0,
- `sqlite`, update from version 3.31.1 to 3.34.0,
- `bc`, update from version 2.7.2 to 3.2.4,
- `patch` and `m4`, bump revision and remove useless file.

I expect this to be the last round of updates for this year, it's been quite productive in my opinion.

PR #70, opened 2020-12-20
Various updates

### LINKED ISSUE #71: ports: xbps configuration is not tailored to managarm
To my knowledge, `xbps` will install a default configuration with the repo pointing at the void linux package repo. This won't work for us, so we should either override this during the build or install our own configuration file with the correct repo url specified.

Dennisbonke:
So it turns out that I was wrong and that pr #68 wasn't the last round of updates for 2020, as I have three more updates. As I'm going on vacation tomorrow, I do expect this to be the last round (for real) unless something major pops up :smile: 

- `libxkbcommon`, updated from version 1.0.2 to 1.0.3, fixes a segfault introduced in 1.0.2 and an issue with keymaps introduced in 0.8.0,
- `xorg-server`, updated from version 1.20.8 to 1.20.10, various improvements all around, and fixes for CVE-2020-14360 and CVE-2020-25712,
- `icu`, updated from version 68.1 to 68.2, a bugfix release, nothing special here,
- `xclock`, add a dependency on `libiconv`,
- `mesa`, updated from version 20.0.5 to 20.3.1,
- `glib`, updated from version 2.66.2 to 2.66.4,
- `harfbuzz`, updated from version 2.7.2 to 2.7.4,
- `libarchive`, updated from version 3.5.0 to 3.5.1,
- `tzdata`, updated from version 2020d to 2020f,
- `gdbm`, updated from version 1.18.1 to 1.19 and dropped a patch,
- `libgpg-error`, updated url, added a patch and updated version from 1.37 to 1.41,
- `xbps`, bumped revision and install some extra data files.

~Blocked on managarm/mlibc#160~
Blocked on managarm/mlibc#180
I think this fixes #71


Dennisbonke:
Holding for `libarchive` version 3.5.1 and `tzdata` version 2020e

avdgrinten:
For libxkbcommon, I added a virtual host-pkg-config so that it reliably finds wayland-scanner. I guess we can do the same for Mesa and drop the hack.

Dennisbonke:
Regarding the `libxkbcommon` patch, sounds good, I’ll try that.

Also holding for `glib` version 2.66.4 and `gdbm` version 1.19.

bootstrap.d/media-libs.yml
```diff
@@ -328,7 +331,7 @@ packages:
         - '--buildtype=debugoptimized'
         #- '--buildtype=debug'
         - '-Dglx=dri'
-        - '-Dplatforms=surfaceless,drm,wayland,x11'
+        - '-Dplatforms=wayland,x11'
```
avdgrinten, line 334: Why do we disable the surfaceless backend?

Dennisbonke: It’s not a valid option anymore, but it is still enabled according to the output at the end of configure

avdgrinten (approved):
LGTM so far.

Dennisbonke:
Alright, should be done. I'm holding off on `SDL2` for now and I've added some `xbps` config files, which I think fixes #71, at least for now (can't test due to missing mlibc support). Can you review it again?

Dennisbonke:
Harfbuzz and tzdata got a bugfix release, nothing major, updated those, pr is still good to merge after managarm/mlibc#180 is merged

patches/libgpg-error/0001-Patch-out-a-broken-operation-during-cross-compilatio.patch
```diff
@@ -0,0 +1,26 @@
...
+diff --git a/src/Makefile.am b/src/Makefile.am
+index fc3acc3..d58871a 100644
+--- a/src/Makefile.am
++++ b/src/Makefile.am
+@@ -353,7 +353,7 @@ gpg-error-config: gpgrt-config gpg-error-config-old
+ 	else                                                   \
+ 	  echo "no";                                           \
+ 	  echo "***  Please report to <https://bugs.gnupg.org> with gpg-error-config-test.log"; \
+-	  exit 1; \
++	  #exit 1; \
```
avdgrinten, line 20: Why is this done?

Dennisbonke: `libgpg-error` tries to execute the newly compiled binary, which fails because it is compiled for managarm. As running it is just a sanity check from `libgpg-error`, I removed the failure out of it so the build succeeds

PR #77, opened 2021-01-19
Add json-c to dev-libs

no92:


bootstrap.d/dev-libs.yml
```diff
@@ -767,3 +767,24 @@ packages:
         environ:
           DESTDIR: '@THIS_COLLECT_DIR@'
         quiet: true
+
+  - name: json-c
+    source:
+      subdir: ports
+      git: 'https://github.com/json-c/json-c'
```
Dennisbonke, line 774: Shouldn't the url be `https://github.com/json-c/json-c.git`?

bootstrap.d/dev-libs.yml
```diff
@@ -767,3 +767,24 @@ packages:
...
+      branch: 'master'
```
Dennisbonke, line 775: We should build from a tag, or if that is not possible, a fixed commit to avoid new commits suddenly breaking our build.
Building from a fixed point also allows for adding version information.

Dennisbonke (changes requested):
In general, we should aim to keep these files alphabetically sorted. Can you move it to the correct spot in the list (below `icu`, line 164).

no92:
Amended the commit to incorporate suggestions.

bootstrap.d/dev-libs.yml
```diff
@@ -162,6 +162,28 @@ packages:
           DESTDIR: '@THIS_COLLECT_DIR@'
         quiet: true
 
+  - name: json-c
+    source:
+      subdir: ports
+      git: 'https://github.com/json-c/json-c.git'
+      branch: 'master'
+      tag: 'json-c-0.15-20200726'
```
Dennisbonke, line 170: Still missing version information, `version: 0.15` just below the `tag` line should do the trick

no92:
Added `version`.

Dennisbonke (approved):

Dennisbonke:
LGTM

PR #78, opened 2021-01-22
sys-apps: Add a port of util-linux

Dennisbonke:
This PR adds a stripped down port of `util-linux` and performs some updates.

- `util-linux`, albeit in a stripped down form, missing various utilities one might expect from this package, like `swapon` or `agetty`. All libraries provided by this package, like `libblkid`, `libfdisk` and `libmount` are available.

- `ncurses`, we now build a seperate `tinfo` library as `util-linux` seems to like that, I'm not quite sure why. Furthermore, we now build the wide character versions of the libraries, making compatibility symlinks to the non-wide version as is common on Linux systems,
- `bash`, added a direct dependency on ncurses and force a rebuild to link against the new ncurses libraries,
- `nano`, override configure on some decisions and force a rebuild to link against the new ncurses libraries,
- `vim`, disable the use of the X windowing system even harder and force a rebuild to link against the new ncurses libraries,
- `less`, force a rebuild to link against the new ncurses libraries.

Blocked on managarm/mlibc#214
Blocked on managarm/mlibc#222


bootstrap.d/sys-libs.yml
```diff
@@ -64,6 +64,7 @@ packages:
       version: '6.1'
     tools_required:
       - system-gcc
+    revision: 3
```
avdgrinten, line 67: Why do we bump from 1 to 3?

patches/util-linux/0001-Stub-out-several-unsupported-things-on-managarm.patch
```diff
@@ -0,0 +1,143 @@
...
+--- a/libmount/src/context_umount.c
++++ b/libmount/src/context_umount.c
+@@ -291,7 +291,9 @@ static int lookup_umount_fs_by_statfs(struct libmnt_context *cxt, const char *tg
+ 		DBG(CXT, ul_debugobj(cxt, "  trying fstatfs()"));
+ 
+ 		/* O_PATH avoids triggering automount points. */
+-		fd = open(tgt, O_PATH);
++		/* MANAGARM: We don't support O_PATH (yet) so disable it. */
++		fd = open(tgt, 0);
++		//fd = open(tgt, O_PATH);
```
avdgrinten, line 124: What happens if we enable this?

patches/util-linux/0002-Revamp-the-configure-script-to-allow-for-more-fine-g.patch
```diff
@@ -0,0 +1,488 @@
+From 629cb91f214d29f55744515577aa5ca898094434 Mon Sep 17 00:00:00 2001
+From: Dennis Bonke <admin@dennisbonke.com>
+Date: Sat, 23 Jan 2021 00:06:46 +0100
+Subject: [PATCH 2/2] Revamp the configure script to allow for more
+ fine-grained control over the programs build. Make configure think that
+ managarm is Linux.
```
avdgrinten, line 6: The commit message of this patch does not have a proper caption.

patches/util-linux/0001-Stub-out-several-unsupported-things-on-managarm.patch
```diff
@@ -0,0 +1,175 @@
...
+ # include <libmount.h>	/* keep it optional for non-linux systems */
+diff --git a/include/namespace.h b/include/namespace.h
+index 2d0a56e02..75d0f439d 100644
+--- a/include/namespace.h
++++ b/include/namespace.h
+@@ -7,6 +7,7 @@
+  */
+ #ifndef UTIL_LINUX_NAMESPACE_H
+ # define UTIL_LINUX_NAMESPACE_H
++#ifndef __managarm__
```
avdgrinten, line 40: What do we patch out here? (And in other places?)

patches/util-linux/0001-Stub-out-several-unsupported-things-on-managarm.patch
```diff
@@ -0,0 +1,175 @@
...
++++ b/term-utils/agetty.c
+@@ -88,6 +88,14 @@
+ #  ifndef DEFAULT_STERM
+ #    define DEFAULT_STERM  "vt102"
+ #  endif
++#elif defined __managarm__
++#  define USE_SYSLOG
++#  ifndef DEFAULT_VCTERM
++#    define DEFAULT_VCTERM "managarm"
++#  endif
```
avdgrinten, line 166: Is this a terminal name? If yes, I don't think we should define a "managarm" terminal type.

avdgrinten (commented):
I have a few comments about the patch to the C / header files. The rest looks good.

patches/util-linux/0001-Stub-out-several-unsupported-things-on-managarm.patch
Dennisbonke: Several linux only features like namespaces or the Linux specific cpu interface

patches/util-linux/0001-Stub-out-several-unsupported-things-on-managarm.patch
Dennisbonke: I think it indeed it. I can change it to match Linux there, that may be more appropriate.

avdgrinten: Agreed.

avdgrinten (approved):
LGTM now.

PR #80, opened 2021-01-24
Add patch for libinput to fix build errors

no92:
This will now use managarm's linux/input.h instead of the in-tree Linux copy, which feels like an ugly hack.

Depends on managarm/mlibc#218


Dennisbonke:
LGTM, waiting on the corresponding mlibc PR to be merged before merging this one.

PR #85, opened 2021-02-03
on debian based systems libssl-dev is required as dependency

sappigeninja:
libssl-dev doesnt come default on ubuntu (tested on lubuntu 20.04).
WIthout it -lnetwork wont link against openssl.

(it just lists libssl-dev as a dependency now in the README)

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ Now proceed to the Building paragraph.
 
 1.  Certain programs are required to build managarm;
     here we list the corresponding Debian packages:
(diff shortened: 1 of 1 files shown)


Dennisbonke:
Can you also add it to the Dockerfile? There is a list there with programs that get installed when one creates a docker image.

sappigeninja:
> Can you also add it to the Dockerfile? There is a list there with programs that get installed when one creates a docker image.

It does seem to be in the Dockerfile.

Dennisbonke:
Hmm. You’re right. Must have missed it then when I looked for it. That also explains why the ci never failed on this. LGTM now.

PR #90, opened 2021-02-21
Deprecate libressl and add OpenSSL

Dennisbonke:
This PR makes the following changes to the ports listed below, all in preparation of removing `libressl`. This decision was made after gentoo announced that they are no longer supporting `libressl` installs because the work required to do so was increasing without the original benefits. We follow them in this decision to lighten the workload.

- `libressl`, deprecate in favor of `openssl`,
- `openssl`, add port,
- `libarchive`, depend on `openssl` instead of `libressl`,
- `python`, depend on `openssl` instead of `libressl` and add a dependency on `util-linux`,
- `cmake`, depend on `openssl` instead of `libressl`,
- `curl`, depend on `openssl` instead of `libressl`,
- `socat`, depend on `openssl` instead of `libressl`,
- `rsync`, depend on `openssl` instead of `libressl`,
- `wget`, depend on `openssl` instead of `libressl` and install config files in `/etc`,
- `xbps`, depend on `openssl` instead of `libressl`.

### DIFF
--- a/bootstrap.d/dev-libs.yml
+++ b/bootstrap.d/dev-libs.yml
@@ -823,6 +823,41 @@ packages:
           DESTDIR: '@THIS_COLLECT_DIR@'
         quiet: true
 
+  - name: openssl
+    source:
+      subdir: ports
+      git: 'https://github.com/openssl/openssl.git'
+      tag: 'OpenSSL_1_1_1j'
+      version: '1.1.1j'
+    tools_required:
+      - system-gcc
+    pkgs_required:
+      - zlib
+    configure:
+      - args:
+        - '@THIS_SOURCE_DIR@/config'
+        - '--prefix=/usr'
+        - '--openssldir=/etc/ssl'
+        - '--libdir=lib'
+        - 'shared'
+        - 'zlib-dynamic'
+        - 'no-afalgeng'
+        environ:
+          CC: 'x86_64-managarm-gcc'
+          CXX: 'x86_64-managarm-g++'
+    build:
+      - args: ['make', '-j@PARALLELISM@']
+      # Disable installing static libraries
+      - args: ['sed', '-i', '/INSTALL_LIBS/s/libcrypto.a libssl.a//', '@THIS_BUILD_DIR@/Makefile']
+      # Suffix all man pages with ssl
+      - args: ['make', 'DESTDIR=@THIS_COLLECT_DIR@', 'MANSUFFIX=ssl', 'install']
+        environ:
+          DESTDIR: '@THIS_COLLECT_DIR@'
+      # Move the doc dir to a versioned directory
+      - args: ['mv', '@THIS_COLLECT_DIR@/usr/share/doc/openssl', '@THIS_COLLECT_DIR@/usr/share/doc/openssl-1.1.1j']
+      # Install more documentation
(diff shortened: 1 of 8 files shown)


bootstrap.d/dev-libs.yml
```diff
@@ -823,6 +823,40 @@ packages:
...
+        - '--libdir=lib'
+        - 'shared'
+        - 'zlib-dynamic'
+        environ:
+          CC: 'x86_64-managarm-gcc'
+          CXX: 'x86_64-managarm-g++'
+    build:
+      - args: ['make', '-j@PARALLELISM@']
+      # DIsable installing static libraries
+      - args: ['sed', '-i', '/INSTALL_LIBS/s/libcrypto.a libssl.a//', '@THIS_BUILD_DIR@/Makefile']
```
avdgrinten, line 850: Why do we do this?

patches/openssl/0001-Add-managarm-support.patch
```diff
@@ -0,0 +1,101 @@
...
+ #define MISALIGN        64
+diff --git a/engines/e_afalg.c b/engines/e_afalg.c
+index 4b17228461..5a445e8154 100644
+--- a/engines/e_afalg.c
++++ b/engines/e_afalg.c
+@@ -21,6 +21,7 @@
+ #include "internal/nelem.h"
+ 
+ #include <sys/socket.h>
++#if !defined (__managarm__)
```
avdgrinten, line 35: What do other OSes do here? I.e., OSes that do not have AFALG?

bootstrap.d/dev-libs.yml
Dennisbonke: Do we want the static libraries? If yes, I’ll remove it. Otherwise, this disables installing it as stated above the sed.

avdgrinten: Installing the static libs doesn't hurt.

avdgrinten: But deps should link against the dynamic ones.

patches/openssl/0001-Add-managarm-support.patch
Dennisbonke: AFALG is Linux specific as far as I can see. So I guess they disable it as well

bootstrap.d/dev-libs.yml
Dennisbonke: Considering we pass disable-static to most packages, I’d say this keeps it consistent with the other packages

ArsenArsen: typo

patches/openssl/0001-Add-managarm-support.patch
avdgrinten: Exactly, that's why I am asking how other OSes disable this. Maybe we should use the same strategy as the other OSes to disable this.

Dennisbonke: I’ve checked haiku, and they don’t explicitly disable it, but they do pass no-asm which may affect it. I’m not aware of any other ports of OpenSSL outside of Linux, the BSD’s and haiku

avdgrinten (approved):
LGTM.

PR #91, opened 2021-03-08
Micro emacs was ported

sappigeninja:
Ported the editor micro emacs.
Having the function Siginterrupt was necessary for the port.

### DIFF
--- a/bootstrap.d/app-editors.yml
+++ b/bootstrap.d/app-editors.yml
@@ -1,4 +1,33 @@
 packages:
+  - name: mg
+    source:
+      subdir: 'ports'
+      git: 'https://github.com/troglobit/mg.git'
+      tag: 'v3.4'
+      version: '3.4'
+      tools_required:
+        - host-autoconf-v2.69
+        - host-automake-v1.15
+        - host-pkg-config
+      regenerate:
+        - args: ['./autogen.sh']
+    tools_required:
+      - host-pkg-config
+      - system-gcc
+    pkgs_required:
+      - ncurses
+    configure:
+      - args:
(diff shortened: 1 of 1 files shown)


bootstrap.d/app-editors.yml
```diff
@@ -1,4 +1,33 @@
...
+      tag: 'v3.4'
+      version: '3.4'
+      tools_required:
+        - host-autoconf-v2.69
+        - host-automake-v1.15
+        - host-pkg-config
+      regenerate:
+        - args: ['./autogen.sh']
+    tools_required:
+      - host-pkg-config
```
Dennisbonke, line 15: Testing suggests that `host-pkg-config` is not required as a tool during configure.

bootstrap.d/app-editors.yml
```diff
@@ -1,4 +1,33 @@
...
+      - system-gcc
+    pkgs_required:
+      - ncurses
+    configure:
+      - args:
+        - '@THIS_SOURCE_DIR@/configure'
+        - '--host=x86_64-managarm'
+        - '--prefix=/usr'
```
Dennisbonke, line 23: Right now, config files get installed into `/usr/etc`, passing `--sysconfdir=/etc` fixes this and places the config files in `/etc`

Dennisbonke (changes requested):
In general, please rebase your changes onto master so that we don't have all this other stuff. Also, does it actually work? Can we edit files with this or do we hit some unimplemented things?

bootstrap.d/app-editors.yml
sappigeninja: Fixing the package
Will reopen when done.

PR #96, opened 2021-04-13
rust: Add cross-compiler port

64:
This PR adds patches for Rust's `std` and `libc` crates and allows us to cross-compile basic Rust programs for managarm.

The full install takes ~1h20 on my machine (4 core low spec laptop), as we compile both stage1 and stage2. This may not be strictly necessary, but is a good idea to avoid ABI issues in future (e.g with procedural macros).

Partially addresses managarm/managarm#91.

### DIFF
--- a/bootstrap.d/dev-lang.yml
+++ b/bootstrap.d/dev-lang.yml
@@ -18,6 +18,18 @@ sources:
     regenerate:
       - args: ['autoreconf', '-f', '-i']
 
+  - name: rust
+    subdir: 'ports'
+    git: 'https://github.com/rust-lang/rust.git'
+    tag: '1.51.0'
+    version: '1.51.0'
+
+  - name: rust-libc
+    subdir: 'ports'
+    git: 'https://github.com/rust-lang/libc.git'
+    tag: '0.2.93'
+    version: '0.2.93'
+
 tools:
   - name: host-python
     from_source: python
@@ -29,6 +41,40 @@ tools:
       - args: ['make', '-j@PARALLELISM@']
     install:
       - args: ['make', 'install']
+
+  - name: host-rust
+    from_source: rust
+    sources_required:
+      - rust-libc
+    tools_required:
+      - host-llvm-toolchain
+    compile:
+      - args: |
+            cat << EOF > config.toml
+            changelog-seen = 2
+
+            [llvm]
+            targets = "X86"
+
+            [build]
+            target = ["x86_64-unknown-managarm-system"]
+            docs = false
+
+            [install]
+            prefix = "@PREFIX@"
+
+            [rust]
+            codegen-tests = false
+
+            [target.x86_64-unknown-linux-gnu]
+            llvm-config = "@BUILD_ROOT@/tools/host-llvm-toolchain/bin/llvm-config"
+
+            [target.x86_64-managarm-system]
+            llvm-config = "@BUILD_ROOT@/tools/host-llvm-toolchain/bin/llvm-config"
+            EOF
+      - args: ['@THIS_SOURCE_DIR@/x.py', 'build', '--stage', '2', '-j', '@PARALLELISM@']
+    install:
+      - args: ['@THIS_SOURCE_DIR@/x.py', 'install', '-j', '@PARALLELISM@']
 
 packages:
   - name: nasm
--- a/bootstrap.d/dev-lang.yml
+++ b/bootstrap.d/dev-lang.yml
@@ -69,7 +69,7 @@ tools:
             [target.x86_64-unknown-linux-gnu]
             llvm-config = "@BUILD_ROOT@/tools/host-llvm-toolchain/bin/llvm-config"
 
-            [target.x86_64-managarm-system]
+            [target.x86_64-unknown-managarm-system]
             llvm-config = "@BUILD_ROOT@/tools/host-llvm-toolchain/bin/llvm-config"
             EOF
       - args: ['@THIS_SOURCE_DIR@/x.py', 'build', '--stage', '2', '-j', '@PARALLELISM@']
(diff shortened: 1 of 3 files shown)


bootstrap.d/dev-lang.yml
```diff
@@ -30,6 +42,40 @@ tools:
...
+            [install]
+            prefix = "@PREFIX@"
+
+            [rust]
+            codegen-tests = false
+
+            [target.x86_64-unknown-linux-gnu]
+            llvm-config = "@BUILD_ROOT@/tools/host-llvm-toolchain/bin/llvm-config"
+
+            [target.x86_64-managarm-system]
```
avdgrinten, line 72: We do not have `-unknown-` here; maybe it should be added for consistency.

bootstrap.d/dev-lang.yml
```diff
@@ -30,6 +42,40 @@ tools:
...
+    compile:
+      - args: |
+            cat << EOF > config.toml
+            changelog-seen = 2
+
+            [llvm]
+            targets = "X86"
+
+            [build]
+            target = ["x86_64-unknown-managarm-system"]
```
avdgrinten, line 60: If `x.py install` doesn't consider this at all, should we maybe just pass `--target` to build instead?

patches/rust-libc/0001-managarm-initial-port.patch
```diff
@@ -0,0 +1,584 @@
...
+ 2 files changed, 553 insertions(+)
+ create mode 100644 src/unix/mlibc/mod.rs
+
+diff --git a/src/unix/mlibc/mod.rs b/src/unix/mlibc/mod.rs
+new file mode 100644
+index 000000000..409ac74f5
+--- /dev/null
++++ b/src/unix/mlibc/mod.rs
+@@ -0,0 +1,550 @@
++// TODO: Should padding be public?
```
avdgrinten, line 18: Hm?

patches/rust-libc/0001-managarm-initial-port.patch
```diff
@@ -0,0 +1,584 @@
...
++
++// Basic data types
++pub type c_char = i8;
++pub type c_long = i64; // TODO ?
++pub type c_ulong = u64; // TODO ?
++pub type wchar_t = ::c_int; // TODO ?
```
avdgrinten, line 24: What are these TODOs?

patches/rust-libc/0001-managarm-initial-port.patch
```diff
@@ -0,0 +1,584 @@
...
++        pub st_mode: mode_t,
++        pub st_nlink: nlink_t,
++        pub st_uid: uid_t,
++        pub st_gid: gid_t,
++        pub st_rdev: dev_t,
++        pub st_size: off_t,
++        // TODO: Should we use timespec here? Don't think we can
++        // pub st_atime: timespec,
++        // pub st_mtime: timespec,
++        // pub st_ctime: timespec,
```
avdgrinten, line 269: Other `std` ports use two members, not a single `timespec` member: https://docs.rs/libc/0.2.93/libc/struct.stat.html (so this comment can probably be deleted).

patches/rust/0001-managarm-initial-port.patch
```diff
@@ -0,0 +1,616 @@
...
+index c3275eb6f0e..8fe5def3ed7 100644
+--- a/library/std/src/sys/unix/thread_local_dtor.rs
++++ b/library/std/src/sys/unix/thread_local_dtor.rs
+@@ -15,7 +15,8 @@
+     target_os = "linux",
+     target_os = "fuchsia",
+     target_os = "redox",
+-    target_os = "emscripten"
++    target_os = "emscripten",
++    target_os = "managarm" // TODO: Is this right?
```
avdgrinten, line 579: Hm?

avdgrinten (commented):
A few comments but overall it looks quite nice.

bootstrap.d/dev-lang.yml
64: It does — what was confusing me was that since `install` builds stage2, the stage1 compiler was being build as linux -> linux (as opposed to `build` which only builds stage1 and thus creates linux -> managarm). So I was stopping the build early, but it does in fact do the right thing when it starts building stage2.

patches/rust-libc/0001-managarm-initial-port.patch
64: Ah yes, I forgot about these. I wasn’t entirely sure what type `wchar` should be, and in future we’ll need to change the definition of long based on the target pointer size. But as we only support x86_64 for now it should be fine

bootstrap.d/dev-lang.yml
avdgrinten: Alright.

patches/rust-libc/0001-managarm-initial-port.patch
64: Yeah, the unix-common code in `libc` relies on that struct being 'inlined' like that. But I wasn't sure if it would change where padding is placed. Seems like it should be fine though, will delete

patches/rust/0001-managarm-initial-port.patch
64: Judge for yourself: https://github.com/rust-lang/rust/blob/07ef25984549bc33adbdc402e977655d8c9093a7/library/std/src/sys/unix/thread_local_dtor.rs#L7-L44

I think we don't need the branch, but I kept it just to be safe. What do you think?

avdgrinten: Yes, that should be fine.

avdgrinten:
LGTM now, let's merge this. Thank you very much for this PR!

PR #97, opened 2021-04-14
update cross files to simplify managarm build

Apache-HB:
this pr supports the changes made in the [meson pr](https://github.com/managarm/managarm/pull/269) made on managarm


scripts/meson-kernel-x86_64-managarm.cross-file
```diff
@@ -5,17 +5,17 @@ ar = 'x86_64-managarm-ar'
 strip = 'x86_64-managarm-strip'
 pkgconfig = 'pkg-config'
 
+[constants]
+# common args that c and cxx need
+args = ['-mno-80387', '-mno-mmx', '-mno-sse', '-mno-sse2', '-ffreestanding']
+# cxx specific arguments
+cxx = args + [ '-target', 'x86_64-managarm-kernel', '-gcc-toolchain', '_BUILD_ROOT_/tools/kernel-gcc']
```
qookei: The C compiler also needs these target arguments, so that when compiling on a host with a different architecture than x86_64 the assembler knows the architecture of .S files.

scripts/meson-clang-common.cross-file
```diff
@@ -0,0 +1,2 @@
+[libarch:project options]
+install_headers = true
```
avdgrinten, line 2: Why is this called `clang-common`? Shouldn't this be something like `system-common` since it applies to `managarm-system`, not `managarm-kernel`?

avdgrinten:
LGTM now. Thank you very much for the PR.

PR #98, opened 2021-04-14
scripts: Add sanity checking to update-image.py

Dennisbonke:
This prevents broken images where the `/bin`, `/lib` and `/sbin` symlinks are not correctly present, thus causing a boot failure


qookei (approved):
LGTM

PR #99, opened 2021-04-20
xbbs: reorganize site setup files

ArsenArsen:


xbbs/x86_64/bootstrap-site.yml
```diff
@@ -4,7 +4,7 @@ define_options:
 
 labels:
     match:
-      - aarch64
+      - x86_64
```
avdgrinten, line 7: x86_64 should not demand any `match:` (all package work for x86_64). The directive should be copied to the aarch64 file, however.

PR #104, opened 2021-05-27
ports: Add neofetch

64:
Requires managarm/managarm#316.


avdgrinten:
Thanks for the port. Merged. :+1:

PR #106, opened 2021-05-28
docker: Revamp the dockerfile

Dennisbonke:
This PR revamps the dockerfile, removing unneeded packages and adding comments explaining why we install the other packages. It also removes xbps from the docker container as it is not used.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -5,63 +5,91 @@ FROM debian:buster AS base
 
 RUN apt-get update \
 	&& apt-get -y install \
+		# Don't install recommended stuff
+		--no-install-recommends \
+		# Used by host-bison
 		autopoint \
+		# Used by binutils in the build process
 		bison \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -5,64 +5,94 @@ FROM debian:buster AS base
...
+                rsync \
+		# Makeinfo is used in gnu build systems
+                texinfo \
+                # Used by protobuf
+                unzip \
+                # We use wget to fetch some stuff further down
+                wget \
+                # Used by eudev
+                xsltproc \
+		# Used by various programs
```
avdgrinten, line 86: Which ones? (At least one example would be nice.)

avdgrinten (commented):
Looks good (except for that one comment) but the indentation is inconsistent.

avdgrinten (commented):
LGTM, thank you!

PR #107, opened 2021-05-31
cbuildrt/rootfs fixes

64:


docker/Dockerfile
```diff
@@ -21,6 +21,7 @@ RUN apt-get update \
 		curl \
 		# Used by binutils in the build process
 		flex \
+		libfl2 \
```
avdgrinten, line 24: Please add a comment why this dependency is needed: `# Required by flex but not a hard dependency in Debian.`

avdgrinten:
LGTM now. Thank you very much!

PR #108, opened 2021-05-31
rust: Fix cargo target selection when setting CARGO_HOME

64:
This PR fixes Rust builds for use with `cbuildrt`.

First, we allow `xbstrap` to set `CARGO_HOME` when building `rustc`. Then, we fix `rustc`'s build system so that it uses the right `--target` for bootstrapping.


avdgrinten (approved):
LGTM. Is that a patch that would be suitable for upstream?

64:
Yes, I'm talking to the `rustc` maintainers a bit on zulip about it now

PR #110, opened 2021-06-02
Fix OpenSSL so it actually works

Dennisbonke:
Due to a faulty configuration, `OpenSSL` had issues with validating certificates. This has been fixed. Furthermore, to better match the Linux configuration, I've remade the configuration patch to look more like it and put it in `10-main`, as opposed to `10-managarm`, as far as I can see there is no need for a separate file here.


patches/openssl/0002-Add-configuration-file-for-Managarm-targets.patch
```diff
@@ -0,0 +1,68 @@
+From d3811a0350dcc9d53d919b5c7b69b92173f1f48e Mon Sep 17 00:00:00 2001
+From: Dennis Bonke <admin@dennisbonke.com>
+Date: Tue, 1 Jun 2021 22:03:10 +0200
+Subject: [PATCH 2/2] Add configuration file for Managarm targets. Initial work
```
ArsenArsen, line 4: this commit message could be formed better

Dennisbonke: Hmmm. I guess I can make the second sentence look more like a proper sentence if that is what you mean, otherwise I’m not really sure what you want me to improve.

ArsenArsen: > openssl: add config file for managarm target

https://chris.beams.io/posts/git-commit/#limit-50
https://chris.beams.io/posts/git-commit/#wrap-72

your editor should assist you with these two basic rules, at least, see `ft=gitcommit` in vim.

avdgrinten: Well, the first line does not fit into 50 chars; I guess that is what Arsen means. We're not really adding a file anymore; we can also just drop "Initial work".

patches/openssl/0002-Add-configuration-file-for-Managarm-targets.patch
```diff
@@ -0,0 +1,68 @@
...
++        inherit_from     => [ "BASE_unix" ],
++        CFLAGS           => picker(default   => "-Wall",
++                         debug       => "-O0 -g",
++                         release     => "-O3"),
++        CXXFLAGS         => picker(default   => "-Wall",
++                         debug       => "-O0 -g",
++                         release     => "-O3"),
++        cxxflags         => add("-std=c++11"),
++        lib_cppflags     => combine("-DOPENSSL_USE_NODELETE", "-DL_ENDIAN"),
++        bn_ops           => "SIXTY_FOUR_BIT_LONG RC4_CHAR",
```
avdgrinten, line 31: Does Linux use this for both x86_64 and aarch64?

Dennisbonke: Seems like it (assuming you're talking about the `bn_ops` thing.

avdgrinten (approved):
LGTM, thank you.

PR #111, opened 2021-06-02
rust: Use rust-patched-libs source group

64:
Requires https://github.com/managarm/xbstrap/pull/49.

This should fix the rust-related CI build failures, finally.

We also move `ripgrep` and `exa` packages to be in alphabetical order.


avdgrinten (approved):
LGTM, thank you for the PR.

PR #112, opened 2021-06-04
Add certificates

Dennisbonke:
This PR adds a port of `make-ca`, which is a script that BLFS uses to generate the certificates used by `OpenSSL`. As the package that holds the certificates is often times called `ca-certificates`, I opted to use the same name. This package has also been added to the `base` metapackage and all dependencies of `ca-certificates` and `ca-certificates` itself received a label marking them ready to build on AArch64.

A small note regarding the version number used on `ca-certificates`, as is common with the `ca-certificates` package, at least on Debian based systems, I have chosen to use the date of package creation / update as a version number. This way, it is clear on which day the certificates have been generated.

### DIFF
--- a/bootstrap.d/app-crypt.yml
+++ b/bootstrap.d/app-crypt.yml
@@ -1,5 +1,6 @@
 packages:
   - name: p11-kit
+    labels: [aarch64]
     source:
       subdir: ports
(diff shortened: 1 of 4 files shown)


avdgrinten (approved):
LGTM, thank you!

PR #118, opened 2021-06-11
bootstrap.yml: only pass -g to image_create for guestfs

64:
Had to lose the comments as it's not very easy to incorporate them with the line wrapping.


avdgrinten (approved):
LGTM, thank you.

PR #121, opened 2021-06-14
bootstrap.yml: add rolling versions for pkgs at 0.0_0

64:


Dennisbonke:
LGTM! Waiting on ci completion before merging it.

PR #122, opened 2021-06-24
Implicit package removal

Dennisbonke:
This PR removes the implicit package definition from `mlibc` and `mlibc-headers`, and adds a dependency on `mlibc` to all packages that require it.


avdgrinten:
LGTM, thank you!

PR #125, opened 2021-07-15
Various updates to xbps

### LINKED ISSUE #71: ports: xbps configuration is not tailored to managarm

Dennisbonke:
This PR updates the `xbps` recipe to no longer install an obsolete file, properly patches the Managarm specific info into the codebase and works around 2 issues in Managarm / mlibc pending their implementation.

This PR is blocked on managarm/mlibc#250, as it contains the buffering fixes and the `lchown` stub that `xbps` wants.
This PR is blocked on managarm/mlibc#288

Properly closes #71

### DIFF
--- a/patches/xbps/0003-Work-around-unimplemented-Managarm-features.patch
+++ b/patches/xbps/0003-Work-around-unimplemented-Managarm-features.patch
@@ -1,0 +1,34 @@
+From 512a7080e61b08eabdc85271ec36da79e326bdcd Mon Sep 17 00:00:00 2001
+From: Dennis Bonke <admin@dennisbonke.com>
+Date: Thu, 15 Jul 2021 02:53:59 +0200
+Subject: [PATCH 3/3] Work around unimplemented Managarm features
+
+Signed-off-by: Dennis Bonke <admin@dennisbonke.com>
+---
+ bin/xbps-install/question.c | 4 ++--
+ 1 file changed, 2 insertions(+), 2 deletions(-)
+
+diff --git a/bin/xbps-install/question.c b/bin/xbps-install/question.c
+index 865649ff..f2d6e9b4 100644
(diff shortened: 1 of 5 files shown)


patches/xbps/0003-Work-around-unimplemented-Managarm-features.patch
```diff
@@ -0,0 +1,64 @@
...
+-	r = strptime(p, "%a, %d %b %Y %H:%M:%S GMT", &tm);
++	//r = strptime(p, "%a, %d %b %Y %H:%M:%S GMT", &tm);
+ 	/*
+ 	 * Some proxies use UTC in response, but it should still be
+ 	 * parsed. RFC2616 states GMT and UTC are exactly equal for HTTP.
+ 	 */
+-	if (r == NULL)
+-		r = strptime(p, "%a, %d %b %Y %H:%M:%S UTC", &tm);
++	//if (r == NULL)
++	//	r = strptime(p, "%a, %d %b %Y %H:%M:%S UTC", &tm);
```
avdgrinten, line 58: This looks a bit strange. What if we just make `strptime` fail in mlibc?

Dennisbonke: Good question, to be fair I haven’t tried that. I’ll try that tomorrow and if that works too I’ll use that.

Dennisbonke: That works too, I'll stub `strptime()` in mlibc and drop this from the patch.

avdgrinten (approved):
LGTM now.

PR #126, opened 2021-07-15
Temporarily add gcc to base as it includes libstdc++ and libgcc

Dennisbonke:
This PR adds `gcc` to `base` and removes it from `base-devel`. This eliminates the need to copy `libgcc_s.so.1` and `libstdc++.so.6` from `system-gcc` as the `gcc` package provides them. This is an temporarily solution, and should be partially reverted whenever `xbstrap` gains the ability to make multiple packages from one build recipe, in which case `gcc` will be split in at least `gcc`, `libgcc` and `libstdc++`, `gcc` will move back to `base-devel` and both `libgcc` and `libstdc++` will be added as a dependency to `base`.

### DIFF
--- a/bootstrap.d/meta-pkgs.yml
+++ b/bootstrap.d/meta-pkgs.yml
@@ -32,6 +32,7 @@ packages:
       - findutils
       - util-linux
       - limine
+      - gcc # Temporary solution to remove the need to copy libstdc++.so and libgcc.so from system-gcc
     configure: []
     build: []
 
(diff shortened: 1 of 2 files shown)


avdgrinten:
We should also adapt the initrd generation script to take the GCC-related `.so` files from the system root, and not from `tools/`.

Dennisbonke:
Good catch, forgot about that one. Will fix tomorrow.

Dennisbonke:
The fix related to the initrd script has been made, and can be found in managarm/managarm#370.

avdgrinten:
LGTM.

PR #127, opened 2021-07-16
Make the meta packages sourceless

Dennisbonke:
The meta packages used to have the same source as managarm, this is unneeded and makes versioning hard to do for these packages. The same goes for the `core-files` package. This PR fixes those packages to be "sourceless" and sets their version to `1.0`.

### DIFF
--- a/bootstrap.d/meta-pkgs.yml
+++ b/bootstrap.d/meta-pkgs.yml
@@ -2,7 +2,9 @@ packages:
   # Meta package with the bare minimum to boot kmscon and be useful
   - name: base
     labels: [aarch64]
-    from_source: managarm
+    source:
+      subdir: meta-sources
+      version: '1.0'
     pkgs_required:
       - mlibc
       - core-files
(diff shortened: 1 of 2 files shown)


bootstrap.d/meta-pkgs.yml
```diff
@@ -2,7 +2,9 @@ packages:
   # Meta package with the bare minimum to boot kmscon and be useful
   - name: base
     labels: [aarch64]
-    from_source: managarm
+    source:
+      subdir: ports
```
avdgrinten, line 6: I wonder if we should put these into `ports` or add a `meta-sources` subdirectory (xbstrap should create such a directory automatically, right?).

Dennisbonke: As far as I know yes it should. Moving it to `meta-sources` sounds good to me, I’ll implement that change tomorrow.

avdgrinten (approved):
LGTM now.

PR #128, opened 2021-07-16
Add a git port

Dennisbonke:
This PR adds an initial port of `git`. Some notable features that are not yet working are pull and clone, but `git log` does work.

~~Blocked on managarm/mlibc#292~~
Blocked on managarm/managarm#482 and managarm/mlibc#750

### DIFF
--- a/patches/git/0001-Initial-Managarm-support.patch
+++ b/patches/git/0001-Initial-Managarm-support.patch
@@ -1,0 +1,35 @@
+From 53584e9beb2c607d6dd77422db652c6014ddabcb Mon Sep 17 00:00:00 2001
+From: Dennis Bonke <admin@dennisbonke.com>
+Date: Fri, 16 Jul 2021 15:58:28 +0200
+Subject: [PATCH] Initial Managarm support
+
+Signed-off-by: Dennis Bonke <admin@dennisbonke.com>
+---
+ config.mak.uname  | 12 ++++++------
+ 1 file changed, 6 insertions(+), 6 deletions(-)
+
+diff --git a/config.mak.uname b/config.mak.uname
+index cb443b4e02..6cb9520c54 100644
+--- a/config.mak.uname
++++ b/config.mak.uname
+@@ -4,12 +4,12 @@
+ # Microsoft's Safe Exception Handling in libraries (such as zlib).
+ # Typically required for VS2013+/32-bit compilation on Vista+ versions.
+ 
+-uname_S := $(shell sh -c 'uname -s 2>/dev/null || echo not')
+-uname_M := $(shell sh -c 'uname -m 2>/dev/null || echo not')
+-uname_O := $(shell sh -c 'uname -o 2>/dev/null || echo not')
+-uname_R := $(shell sh -c 'uname -r 2>/dev/null || echo not')
+-uname_P := $(shell sh -c 'uname -p 2>/dev/null || echo not')
+-uname_V := $(shell sh -c 'uname -v 2>/dev/null || echo not')
++uname_S := #$(shell sh -c 'uname -s 2>/dev/null || echo not')
++uname_M := #$(shell sh -c 'uname -m 2>/dev/null || echo not')
++uname_O := #$(shell sh -c 'uname -o 2>/dev/null || echo not')
++uname_R := #$(shell sh -c 'uname -r 2>/dev/null || echo not')
++uname_P := #$(shell sh -c 'uname -p 2>/dev/null || echo not')
++uname_V := #$(shell sh -c 'uname -v 2>/dev/null || echo not')
+ 
+ ifdef MSVC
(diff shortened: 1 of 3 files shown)


patches/git/0001-Initial-Managarm-support.patch
```diff
@@ -0,0 +1,58 @@
...
+-static inline void flockfile(FILE *fh)
+-{
+-	; /* nothing */
+-}
+-static inline void funlockfile(FILE *fh)
+-{
+-	; /* nothing */
+-}
++#define flockfile(fh)
++#define funlockfile(fh)
```
avdgrinten, line 52: Why is this done?

avdgrinten:
Do commands on local repositories work? (For example after copying a repository into the sysroot?) Stuff like `git add`, `git status`, `git commit`?

Dennisbonke:
I haven’t tried `git add` recently so I can’t tell you that, I know that `git log` and `git diff` seem to work and from what I remember was `git status` broken at this time.

patches/git/0001-Initial-Managarm-support.patch
Dennisbonke: With the original code, I got build failures, something about non static declarations if I remember correctly. In an old patch i fixed that like this, but that is so long ago I’m not quite sure why I opted for that approach.

Dennisbonke:
I just checked, `git status` is broken on the lack of an `setitimer()` implementation, `git diff` works (albeit slow), `git log` works fine, `git add` seems to work fine, `git commit` works too (after setting the identity with `git config`), `git reset --hard` fails on `setitimer()` too and so does `git checkout HEAD~1`. Anything else you want me to test before merging?

patches/git/0001-Initial-Managarm-support.patch
avdgrinten: The removed code shoud be a no-op already. Try removing this part of the patch and try to resolve errors if they occur.

patches/git/0001-Initial-Managarm-support.patch
```diff
@@ -0,0 +1,35 @@
...
+-uname_O := $(shell sh -c 'uname -o 2>/dev/null || echo not')
+-uname_R := $(shell sh -c 'uname -r 2>/dev/null || echo not')
+-uname_P := $(shell sh -c 'uname -p 2>/dev/null || echo not')
+-uname_V := $(shell sh -c 'uname -v 2>/dev/null || echo not')
++uname_S := #$(shell sh -c 'uname -s 2>/dev/null || echo not')
++uname_M := #$(shell sh -c 'uname -m 2>/dev/null || echo not')
++uname_O := #$(shell sh -c 'uname -o 2>/dev/null || echo not')
++uname_R := #$(shell sh -c 'uname -r 2>/dev/null || echo not')
++uname_P := #$(shell sh -c 'uname -p 2>/dev/null || echo not')
++uname_V := #$(shell sh -c 'uname -v 2>/dev/null || echo not')
```
avdgrinten, line 30: Why is this done?

Dennisbonke: If left as it was, the uname will match Linux, and we don’t build for Linux. In my memory serenity used to patch it like this (and they did), but they removed this patch a few months ago.

Dennisbonke:
It seems like serenity patches out several calls to setitimer, which we might be able to use to get more commands working.

patches/git/0001-Initial-Managarm-support.patch
avdgrinten: Does this run on the host? Where does the uname show up when running git?

avdgrinten:
I'm not a fan of doing that; we should just implement POSIX timers.

patches/git/0001-Initial-Managarm-support.patch
Dennisbonke: Serenity reinstated the patch not too long ago, see [this issue](https://github.com/SerenityOS/serenity/issues/9357) and the [linked pr](https://github.com/SerenityOS/serenity/pull/9377) for an explanation why.

Dennisbonke:
I agree on the not a fan part, however, do we want to block this port on that? If yes, then I'll mark this one as a draft waiting on POSIX timers.

patches/git/0001-Initial-Managarm-support.patch
ArsenArsen: these can be overwritten on the make command line regardless of the state of the makefile

Dennisbonke:
> I agree on the not a fan part, however, do we want to block this port on that? If yes, then I'll mark this one as a draft waiting on POSIX timers.

This has been resolved, `setitimer` and friends are sysdeps which Managarm doesn't implement. `git` is fine with that.

Dennisbonke:
Superseeded by #510

PR #129, opened 2021-07-22
util-linux: Split into a library and binaries package

Dennisbonke:
This PR changes `util-linux` to produce `util-linux-libs`, which contains the libraries and header files that `util-linux` provides, and `util-linux`, which holds binaries and their documentation. The rationale behind this change is allowing `eudev` to be build with `libblkid` support (to be enabled at a later date), which is currently impossible as `util-linux` uses `eudev` itself when building certain programs, creating a circular dependency.


avdgrinten (approved):
LGTM.

Issue #103, opened 2021-05-09
Intermittent failure of `update_image.py` with Docker mount: cannot create regular file `/mnt/boot/efi`


64:
Sometimes when running `xbstrap run make-image` using Docker I get the following error:

```
$ xbstrap run make-image qemu
xbstrap: Running the following plan:
    run            update-initrd
    run            make-image
    run            qemu
xbstrap: run update-initrd [1/3]
xbstrap: Running ['@SOURCE_ROOT@/managarm/tools/gen-initrd.py', '--triple', '@OPTION:arch-triple@'] (tools: ['cross-binutils']) in Docker
xbstrap: run make-image [2/3]
xbstrap: Running ['@SOURCE_ROOT@/scripts/update-image.py', '--triple', '@OPTION:arch-triple@', '--mount-using', '@OPTION:mount-using@'] (tools: [])
update-image: Running the following plan:
 - mount
 - update file system
 - unmount
update-image: Running action mount ("image" to "mountpoint" using docker)
update-image: Running action update file system (from "system-root" to "mountpoint")
update-image: Updating the image doesn't require root
update-image: Update target has an EFI directory
update-image: Updating the file system...
+ mkdir -p /mnt/boot/managarm
+ install system-root/usr/managarm/bin/eir-stivale /mnt/boot/managarm
+ install -s --strip-program=tools/cross-binutils/bin/x86_64-managarm-strip system-root/usr/managarm/bin/eir-mb1 /mnt/boot/managarm
+ install -s --strip-program=tools/cross-binutils/bin/x86_64-managarm-strip system-root/usr/managarm/bin/thor /mnt/boot/managarm
+ install initrd.cpio /mnt/boot/managarm
+ mkdir -p /mnt/root
+ mkdir -p /mnt/usr/bin
+ mkdir -p /mnt/usr/lib
+ mkdir -p /mnt/var
+ mkdir -p /mnt/dev
+ mkdir -p /mnt/proc
+ mkdir -p /mnt/run
+ mkdir -p /mnt/sys
+ mkdir -p /mnt/tmp
+ mkdir -p /mnt/boot/grub
+ mkdir -p /mnt/home
+ cp /stuff/scripts/grub.cfg /mnt/boot/grub
+ cp /stuff/scripts/limine.cfg /mnt/boot/limine.cfg
[...]

PR #130, opened 2021-07-23
scripts/update-image: fix lint errors

64:
Fix lint errors according to `pyright`.

Possibly related to managarm/bootstrap-managarm#103.

### DIFF
--- a/scripts/update-image.py
+++ b/scripts/update-image.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-import argparse, attr, shutil, shlex, subprocess, sys, os, errno, json, time
+import argparse, shutil, shlex, subprocess, sys, os, errno, json, time
 
 elevation_method = None
 verbose = False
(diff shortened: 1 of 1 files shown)


scripts/update-image.py
```diff
@@ -9,21 +9,21 @@
...
-@attr.s
+	def __init__(self, blockdev, mount_using, image, mountpoint, root_idx, efi_idx, root_uuid):
+		self.blockdev = blockdev
+		self.mount_using = mount_using
+		self.image = image
+		self.mountpoint = mountpoint
+		self.root_idx = root_idx
+		self.efi_idx = efi_idx
+		self.root_uuid = root_uuid
+
```
ArsenArsen, line 21: No, attr.ib is used for a reason.

scripts/update-image.py
```diff
@@ -9,21 +9,21 @@
...
 class Partition:
-	idx = attr.ib()
-	type = attr.ib()
-	uuid = attr.ib()
+	def __init__(self, idx, type, uuid):
+		self.idx = idx
+		self.type = type
+		self.uuid = uuid
```
ArsenArsen, line 26: Likewise (is this missing an `@attr.s`?)

scripts/update-image.py
```diff
@@ -299,9 +300,9 @@ def run(self):
 			# Determine root uuid manually for mountpoint
 			diskdev, partdev = dev_for_mountpoint(self.mountpoint)
 
-			if os.isdir('/dev/disk/by-uuid'):
+			if os.path.isdir('/dev/disk/by-uuid'):
```
ArsenArsen, line 303: `import os.path as path` is a common convention.

ArsenArsen (changes requested):
Furthermore, rather than these faulty lints we should consider applying `flake8` and some plugins to all our Python code, to enforce PEP8 guidelines and some other QA (import order, complexity requirements).

scripts/update-image.py
64: I get that. But AFAICT the only feature we use is __init__. So I've just removed attrs entirely from this script.

64:
I agree, we should really use PEP8.

scripts/update-image.py
64: I can replace all instances of `os.path` -> `path`, if that's what you're suggesting?

avdgrinten:
I'm fine with keeping `os.path`. I'm indifferent about `attr.s`; if something can be achieved without an external lib, that's probably better anyway. @ArsenArsen what is the real advantage of `attr.s` here, other than not having to write the constructor?

ArsenArsen:
Here? Nothing specifically, this just seemed like a naive lint. If the dep was there already, and it'd appear that it was, it's probably fine to keep it, though.

As for `path`, I'm just mentioning a widespread convention.

avdgrinten:
Why did the linter complain about `attr.s`?

ArsenArsen:
Don't know, I'm just going off the initial comment:

> Fix lint errors according to `pyright`.

64:
> Why did the linter complain about `attr.s`?

To be clear: it's definitely a false positive lint. The decorator seems to confuse it. There's nothing actually broken about attrs here, unlike the other lint errors.

avdgrinten:
Alright, let's merge this now since there are no hard issues.

PR #131, opened 2021-07-24
Add workflow to build from xbps repos

64:
I also snuck in a commit to only upload rootfs from source repository. This prevents CI failing on forks.


avdgrinten (approved):
LGTM, thank you!

PR #133, opened 2021-08-01
rust: install packages to /usr/bin rather than /bin

64:
Otherwise this corrupts the sysroot.

This wasn't actually a noticeable problem before we started using `xbps`.


Dennisbonke:
LGTM

Issue #39, opened 2020-06-07
Missing ports for partial self-hosting


Dennisbonke:
For managarm to be partially self hosting, meaning that we can compile managarm-kernel and managarm-system on managarm with the sources patched and available from the host, so that we only need to configure and build them, the following packages are needed (possible incomplete list):

- [ ] `meson`, managarm's primary build system is `meson`.
- [ ] `ninja`, managarm's primary builder is `ninja`.
- [ ] `xbstrap`, managarm's master build system is executed by `xbstrap`, this needs some additional python modules as well.
- [ ] `pyyaml`, `xbstrap` depends on `pyyaml`.
- [ ] `colorama`, `xbstrap` depends on `colorama`.
- [ ] `llvm + clang`, while we do build `llvm`, we do not build `clang`, the kernel is compiled with clang, and as such, this is needed.

Furthermore, the following ports require extra attention:

- [ ] `binutils`, currently works, but when called from a makefile `as` crashes
- [ ] `gcc`, `gcc` works, but invoking `g++` fails, furthermore, calling `gcc` as `cc` (a symlink mandated by the FHS, and used for compatibility reasons in many build systems), crashes in mlibc (see issue managarm/mlibc#93) (This has been fixed in the meantime)
- [ ] `python`, more modules need to be available, and several python packages need to be ported
- [ ] `pkg-config`, fails when running, complains about `ungetc()`

PR #138, opened 2021-08-12
Add ports of ninja and tcl

Dennisbonke:
This PR adds two new ports, `ninja` and `tcl`, both in various states of working. `ninja` requires `sigpending` to be useful and `tcl` is a user of `pthread_atfork` (see managarm/mlibc#298). Furthermore, ninja requires some mlibc work in `fopen`, a PR for that will be posted soon.

Part of #39 
Complemented (but not blocked!) by managarm/mlibc#299

### DIFF
--- a/bootstrap.d/dev-util.yml
+++ b/bootstrap.d/dev-util.yml
@@ -150,6 +150,35 @@ packages:
         environ:
           DESTDIR: '@THIS_COLLECT_DIR@'
 
+  - name: ninja
+    architecture: '@OPTION:arch@'
+    metadata:
+      summary: The Ninja build system
+      description: Ninja is a small build system similar to make.
+      spdx: 'Apache-2.0'
+      website: 'https://ninja-build.org/'
+      maintainer: "Dennis Bonke <dennis@managarm.org>"
+      categories: ['dev-util']
+    source:
+      subdir: ports
+      git: 'https://github.com/ninja-build/ninja.git'
+      tag: 'v1.10.2'
+      version: '1.10.2'
+    tools_required:
+      - system-gcc
+      - host-cmake
+    configure:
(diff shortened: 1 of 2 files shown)


bootstrap.d/dev-util.yml
```diff
@@ -150,6 +150,35 @@ packages:
...
+      - host-cmake
+    configure:
+      - args:
+        - 'cmake'
+        - '-DCMAKE_TOOLCHAIN_FILE=@SOURCE_ROOT@/scripts/CMakeToolchain-@OPTION:arch-triple@.txt'
+        - '@THIS_SOURCE_DIR@'
+    build:
+      - args: ['make', '-j@PARALLELISM@']
+      - args: ['mkdir', '-p', '@THIS_COLLECT_DIR@/usr/bin']
+      - args: ['cp', 'ninja', '@THIS_COLLECT_DIR@/usr/bin']
```
64, line 178: Is there no 'make install' rule?

64: I just checked on a local clone with `make install -n` and it seems to do *something*, I think it should exist. Can you try?

ArsenArsen: [Gentoo seems to agree.](https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-util/ninja/ninja-9999.ebuild?id=f2ba94c014da24115072b0fdfd2ddea74c8ed81e#n104)

Dennisbonke: @64 is right here, make install seems to work as intended. Switched to that, thanks for catching it.

Dennisbonke:
As @avdgrinten approved these changes on Discord, I’ll merge this PR.

PR #139, opened 2021-08-30
Add a bunch of new ports, working towards WebKitGTK

Dennisbonke:
This PR adds a bunch of new ports, all in preparation for upstreaming `gtk+3` and `WebKitGTK`. More ports are expected, but are in need of a cleanup. Furthermore, some post install scripts have been added and the Dockerfile has been updated to accommodate the new ports.

New ports:
- `libgcrypt`,
- `libunistring`,
- `libidn2`,
- `gnutls`,
- `gsettings-desktop-schemas`, includes a post install script to compile the schemas.
- `glib-networking`,
- `libpsl`,
- `libsoup`,
- `xcursor-themes`,
- `hicolor-icon-theme`.

New tools:
- `host-gobject-introspection`, only used for some aclocal files.

Updated ports:
- `shared-mime-info`, added a post install script to build the mime database,
- `gtk+2`, rename a file which conflicts with `gtk+3`, add a post install script to build a cache and depend on `hicolor-icon-themes`.

### DIFF
--- a/bootstrap.d/net-dns.yml
+++ b/bootstrap.d/net-dns.yml
@@ -1,0 +1,72 @@
+packages:
+  - name: libidn2
+    architecture: '@OPTION:arch@'
+    metadata:
+      summary: An implementation of the IDNA2008 specifications (RFCs 5890, 5891, 5892, 5893)
+      description: This package provides a library for internationalized string handling based on standards from the Internet Engineering Task Force (IETF)'s IDN working group, designed for internationalized domain names.
+      spdx: 'LGPL-3.0-or-later'
+      website: 'https://www.gnu.org/software/libidn/#libidn2'
+      maintainer: "Dennis Bonke <dennis@managarm.org>"
+      categories: ['net-dns']
+    source:
+      subdir: 'ports'
+      git: 'https://gitlab.com/libidn/libidn2.git'
+      tag: 'v2.3.2'
+      version: '2.3.2'
+      tools_required:
+        - host-autoconf-v2.69
+        - host-automake-v1.15
+        - host-libtool
+        - host-pkg-config
+      regenerate:
+        - args: ['./bootstrap']
+        - args: ['autoreconf', '-fvi']
+    tools_required:
+      - system-gcc
+      - host-autoconf-v2.69
+      - host-automake-v1.15
+      - host-libtool
+      - host-pkg-config
+    pkgs_required:
+      - mlibc
+      - libunistring
+      - libiconv
+    configure:
+      # Remove some files from the source directory if they exist.
+      - args: |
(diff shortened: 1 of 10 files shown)


bootstrap.d/x11-libs.yml
```diff
@@ -174,6 +176,12 @@ packages:
         environ:
           DESTDIR: '@THIS_COLLECT_DIR@'
         quiet: true
+      # Rename a file to avoid conflicts, we will use gtk3 as default, so rename this one.
+      - args: ['mv', '-v', '@THIS_COLLECT_DIR@/usr/bin/gtk-update-icon-cache', '@THIS_COLLECT_DIR@/usr/bin/gtk2-update-icon-cache']
+    scripts:
+        post_install:
+          - args: 'echo "Running update-mime-database; this may a few moments..."'
+          - args: ['gtk-query-immodules-2.0 --update-cache']
```
avdgrinten, line 184: The message does not match the command.

bootstrap.d/net-dns.yml
```diff
@@ -0,0 +1,72 @@
...
+    configure:
+      # Remove some files from the source directory if they exist.
+      - args: |
+              if [ -f @THIS_SOURCE_DIR@/lib/gendata ]; then
+                  rm -rv @THIS_SOURCE_DIR@/lib/gendata
+              fi
+      - args: |
+              if [ -f @THIS_SOURCE_DIR@/lib/gentr46map ]; then
+                  rm -rv @THIS_SOURCE_DIR@/lib/gentr46map
+              fi
```
avdgrinten, line 43: Why is this done?

bootstrap.d/x11-libs.yml
Dennisbonke: Oops, I'll fix that in a bit.

bootstrap.d/net-dns.yml
Dennisbonke: In a clean source, these shouldn't exist. If they do exist, they can cause issues on rebuilds.

Dennisbonke:
> Does `update-mine-database` really take that long?

Yeah, it is quite slow, even on modern Linux systems to my knowledge.

avdgrinten:
LGTM, but before we add `update-mime-database` as a dependency of the default image (on builds.managarm.org), we need to fix the boot time on first boot (i.e., when the hook is triggered).

Dennisbonke:
> LGTM, but before we add `update-mime-database` as a dependency of the default image (on builds.managarm.org), we need to fix the boot time on first boot (i.e., when the hook is triggered).

Agreed, but the update process is just slow, even on Linux. As discussed on Discord, running these first boot hooks during image generation on the CI might be the solution. Merging this as this PR does not add `update-mime-database` to the default image.

PR #142, opened 2021-09-23
Add a cups port

Dennisbonke:
This PR adds a port of `cups` and updates the `passwd` and `group` files to accommodate for this, by adding the `lpadmin` group and the `lp` user.

### DIFF
--- a/bootstrap.d/net-print.yml
+++ b/bootstrap.d/net-print.yml
@@ -1,0 +1,56 @@
+packages:
+  - name: cups
+    metadata:
+      summary: The Common Unix Printing System
+      description: The Common Unix Printing System (CUPS) is a print spooler and associated utilities. It is based on the "Internet Printing Protocol" and provides printing services to most PostScript and raster printers.
+      spdx: 'Apache-2.0'
+      website: 'https://www.cups.org/'
+      maintainer: "Dennis Bonke <dennis@managarm.org>"
+      categories: ['net-print']
+    source:
+      subdir: ports
+      git: 'https://github.com/apple/cups.git'
+      tag: 'v2.3.3'
+      version: '2.3.3'
+      tools_required:
+        - host-autoconf-v2.69
+        - host-automake-v1.15
+      regenerate:
+        - args: ['cp',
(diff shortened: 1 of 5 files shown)


bootstrap.d/net-print.yml
```diff
@@ -0,0 +1,56 @@
...
+        - '--with-system-groups=lpadmin'
+        - '--with-docdir=/usr/share/cups/doc-2.3.3op2'
+        - '--disable-dbus'
+        - '--disable-gssapi'
+        - '--disable-pam'
+        - '--disable-avahi'
+        - '--disable-dnssd'
+        - '--disable-browsing'
+        - '--with-components=libcups'
+      - args: 'sed -i s/-Werror//g @THIS_BUILD_DIR@/Makedefs'
```
avdgrinten, line 49: The configure script does not allow disabling `-Werror`? Likewise, patching the source does not work?

extrafiles/group
```diff
@@ -2,9 +2,10 @@ root:x:0:
 bin:x:1:daemon
 tty:x:5:
 disk:x:6:
-lp:x:7:
 daemon:x:8:
+lp:x:9:
```
avdgrinten, line 6: Why is the ID changed here? Where does it come from?

bootstrap.d/net-print.yml
Dennisbonke: Unfortunately not, even gentoo applies a sed to fix this issue.

extrafiles/group
Dennisbonke: This brings it in line with LFS/BLFS, which it shoud've been in the first place, as that was my reference. Probably a mistake during the initial generation of the file.

avdgrinten: Add a comment to the top of the file that user IDs should be in sync with BLFS (where possible).

Dennisbonke: `/etc/passwd` and `/etc/group` do not allow for comments unfortunately.

avdgrinten (approved):
LGTM.

PR #143, opened 2021-09-23
A few new ports

Dennisbonke:
This PR adds 4 new ports that are listed as required or recommended dependencies of `OpenJDK` according to BLFS. Even if not needed for that, 3 of these ports are general nice to haves.

The following new ports were added:
- `cpio`,
- `giflib`,
- `unzip`,
- `zip`.

### DIFF
--- a/bootstrap.d/app-arch.yml
+++ b/bootstrap.d/app-arch.yml
@@ -34,6 +34,48 @@ packages:
       - args: ['ln', '-sf', 'bzgrep', '@THIS_COLLECT_DIR@/usr/bin/bzfgrep']
       - args: ['ln', '-sf', 'bzmore', '@THIS_COLLECT_DIR@/usr/bin/bzless']
 
+  - name: cpio
+    architecture: '@OPTION:arch@'
+    metadata:
+      summary: The cpio package contains tools for archiving
+      description: A file archival tool which can also read and write tar files.
+      spdx: 'GPL-3.0-only'
+      website: 'https://www.gnu.org/software/cpio/cpio.html'
+      maintainer: "Dennis Bonke <dennis@managarm.org>"
+      categories: ['app-arch']
+    source:
+      subdir: ports
+      git: 'https://git.savannah.gnu.org/git/cpio.git'
+      tag: 'release_2_13'
+      version: '2.13'
+      tools_required:
+        - host-autoconf-v2.69
+        - host-automake-v1.15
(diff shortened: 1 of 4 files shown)


bootstrap.d/app-arch.yml
```diff
@@ -190,6 +232,50 @@ packages:
...
+      patch-path-strip: 1
+      version: '6.0'
+    tools_required:
+      - system-gcc
+    pkgs_required:
+      - mlibc
+      - bzip2
+      - libiconv
+    configure:
+      # Fixup their shitty makefile, thanks gentoo!
```
avdgrinten, line 258: Remove the swear word / let's not blame other packages here.

avdgrinten (commented):
Why does the unzip package come with a 16K patch file?

Dennisbonke:
> Why does the unzip package come with a 16K patch file?

Courtesy of BLFS (which I should probably mention). It fixes a bunch of CVE's (listed in the patch) and some other issues. The last release of unzip has been a while. Also, from what I saw, the patch includes the unmodified source files as <name.c.orig>, inflating the size.

bootstrap.d/app-arch.yml
Dennisbonke: You're right, my bad. Fixed it, still thanking gentoo for the sed as it is from there.

avdgrinten:
This PR still has the 16K sloc patch file. Is that taken as-is from a reputable source? If yes, we should probably link to that source. Otherwise I'd be hesitant to merge this since we cannot validate the 16K sloc patch.

Dennisbonke:
> This PR still has the 16K sloc patch file. Is that taken as-is from a reputable source? If yes, we should probably link to that source. Otherwise I'd be hesitant to merge this since we cannot validate the 16K sloc patch.

It was taken from [here](https://www.linuxfromscratch.org/blfs/view/svn/general/unzip.html), with a direct [download link](https://www.linuxfromscratch.org/patches/blfs/svn/unzip-6.0-consolidated_fixes-1.patch) available too. The submitter used an at linuxfromscratch.org and the patch was based on Fedora patches, as the patch itself notes. I deem this source trustable.

avdgrinten:
Please add a link to the source to the patch's commit message.

Dennisbonke:
> Please add a link to the source to the patch's commit message.

Done as requested.

PR #144, opened 2021-10-01
docker: Add 2 dependencies for swi-prolog

Dennisbonke:
This PR adds `libncurses-dev` and `libreadline-dev` to the rootfs, which seems to be required to build `swi-prolog`. The resulting rootfs grows from `789MB` to `792MB`.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -51,11 +51,14 @@ RUN apt-get update \
 		libgmp-dev \
 		libmpc-dev \
 		libmpfr-dev \
+		# Used by ace
+		libpng-dev \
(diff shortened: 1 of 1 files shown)


ArsenArsen:
why wasn't this merged?

avdgrinten:
I'm not sure if we shouldn't build these ourselves, especially if we already build them for Managarm, too.

Dennisbonke:
We’re not sure if this is the right way to go, also the port that this was for has stalled out and hasn’t received progress in a few months. We might be better off closing this and reopening it if needed or something like that.

PR #146, opened 2021-11-25
Add dsda-doom port

qookei:


bootstrap.d/games-fps.yml
```diff
@@ -0,0 +1,110 @@
...
+      - args: ['ninja', 'install']
+        environ:
+          DESTDIR: '@THIS_COLLECT_DIR@'
+        quiet: true
+
+  - name: tyr-quake
+    architecture: '@OPTION:arch@'
+    source:
+      subdir: 'ports'
+      git: 'git://disenchant.net/tyrquake'
```
Dennisbonke, line 78: Don’t we want to use https here too?

Dennisbonke (commented):
Looks good, one comment.

bootstrap.d/games-fps.yml
qookei: This is the link given on the website, and cloning from `https://disenchant.net/git/tyrquake.git` (which uses the dumb http transport) is significantly slower than cloning from `git://disenchant.net/tyrquake` (2min vs 7s)

ArsenArsen: doesn't matter, no need to use a nontls protocol

qookei: Hm, good point

ArsenArsen: ... or https://disenchant.net/files/engine/tyrquake-0.69.tar.gz. Though, we lack c-sum checking for now. I'll put that on my xbstrap todo list (or on the issue tracker).

Dennisbonke: Yeah after reading this convo I think we want to use https here too, especially since mirroring is a thing. If you change to that I can approve and merge this PR.

avdgrinten: What's the advantage of https over git? Why should we switch even if qookie tried it out and it's much slower?

avdgrinten: Is the git protocol unencrypted by default? I thought git meant "git over SSH" but maybe I'm wrong here.

qookei: Looks like it's unencrypted:
```
connect(3, {sa_family=AF_INET, sin_port=htons(9418), sin_addr=inet_addr("112.213.32.136")}, 16) = 0
setsockopt(3, SOL_SOCKET, SO_KEEPALIVE, [1], 4) = 0
dup(3)                                  = 4
write(4, "003d", 4)                     = 4
write(4, "git-upload-pack /tyrquake\0host=d"..., 57) = 57
```

ArsenArsen: Nope, `git://` is git's smart protocol over plain TCP. Here's an example packet:

```
0000   30 30 33 64 67 69 74 2d 75 70 6c 6f 61 64 2d 70   003dgit-upload-p
0010   61 63 6b 20 2f 74 79 72 71 75 61 6b 65 00 68 6f   ack /tyrquake.ho
0020   73 74 3d 64 69 73 65 6e 63 68 61 6e 74 2e 6e 65   st=disenchant.ne
0030   74 00 00 76 65 72 73 69 6f 6e 3d 32 00            t..version=2.
```

I usually see a slowdown of about 5-10% max (which is probably attributable to TLS overhead) but in this case it's over 15x slower. I assume the `https://` server in this case isn't a `git-http-backend` or such.

qookei: One slight problem, xbstrap can't clone over the dumb transport since it needs shallow clones:
```
xbstrap: fetch tyr-quake [1/8]
fatal: dumb http transport does not support shallow capabilities
xbstrap: Action fetch of source tyr-quake failed
```

avdgrinten: There is a `no_shallow_fetch: true` property to work around this issue.

bootstrap.d/games-fps.yml
```diff
@@ -0,0 +1,111 @@
...
+    build:
+      - args: ['ninja']
+      - args: ['ninja', 'install']
+        environ:
+          DESTDIR: '@THIS_COLLECT_DIR@'
+        quiet: true
+
+  - name: tyr-quake
+    architecture: '@OPTION:arch@'
+    source:
```
Dennisbonke, line 75: Small nitpick, metadata above source declaration is the common way to put this, For consistency, we should do the same here.

Dennisbonke (commented):
One really small nitpick, then it can be merged.

Dennisbonke (approved):
LGTM

PR #147, opened 2021-11-27
core-files: Add missing directories and initialize them

Dennisbonke:
This PR fixes some strange errors during boot relating to mount failures.


avdgrinten (commented):
Why is /boot/grub created unconditionally? Shouldn't grub create that at installation time? (And if GRUB is not installed, it should not be needed.)

Dennisbonke:
I've just replicated the directories from update-image.py, it seems to be created unconditionally there too.

bootstrap.yml
```diff
@@ -678,19 +678,36 @@ packages:
...
+      # xbps does not install empty directories, .keep files fixes this
+      - args: ['touch', '@THIS_COLLECT_DIR@/etc/.keep']
+      - args: ['touch', '@THIS_COLLECT_DIR@/run/.keep']
+      - args: ['touch', '@THIS_COLLECT_DIR@/root/.keep']
+      - args: ['touch', '@THIS_COLLECT_DIR@/dev/.keep']
+      - args: ['touch', '@THIS_COLLECT_DIR@/sys/.keep']
+      - args: ['touch', '@THIS_COLLECT_DIR@/proc/.keep']
+      - args: ['touch', '@THIS_COLLECT_DIR@/tmp/.keep']
+      - args: ['touch', '@THIS_COLLECT_DIR@/boot/grub/.keep']
+      - args: ['touch', '@THIS_COLLECT_DIR@/home/managarm/.keep']
```
ArsenArsen, line 709: likewise

bootstrap.yml
```diff
@@ -678,19 +678,36 @@ packages:
...
       # Create initial directories
       - args: ['mkdir', '-p', '@THIS_COLLECT_DIR@/etc']
       - args: ['mkdir', '-p', '@THIS_COLLECT_DIR@/run']
       - args: ['mkdir', '-p', '@THIS_COLLECT_DIR@/root']
+      - args: ['mkdir', '-p', '@THIS_COLLECT_DIR@/dev']
+      - args: ['mkdir', '-p', '@THIS_COLLECT_DIR@/sys']
+      - args: ['mkdir', '-p', '@THIS_COLLECT_DIR@/proc']
+      - args: ['mkdir', '-p', '@THIS_COLLECT_DIR@/tmp']
+      - args: ['mkdir', '-p', '@THIS_COLLECT_DIR@/boot/grub']
       - args: ['mkdir', '-p', '@THIS_COLLECT_DIR@/home/managarm']
```
ArsenArsen, line 693: out of scope for packages

ArsenArsen (changes requested):
I'm not sure how to handle home and such. Maybe making home is fine while we hardcode passwd?

avdgrinten:
We can probably keep hardcoding /home/managarm for now.

Dennisbonke:
As requested, the grub directories are no longer there. As the consensus is that we can continue hardcoding `/home/managarm` for now, this should be good to go now.

ArsenArsen:
sure, but no packages should contain anything in /home, especially considering we'll remove it.
it's better to create/remove this dir when making an image IMHO.

Dennisbonke:
> sure, but no packages should contain anything in /home, especially considering we'll remove it. it's better to create/remove this dir when making an image IMHO.

Eventually we should probably remove that yeah, instead putting common dotfiles in `/etc/skel` and letting `useradd` handle it. For now, this should be fine tho.

avdgrinten (approved):

PR #149, opened 2021-12-04
scripts/[risc]: move needs_exe_wrapper

ArsenArsen:
this option is for [properties] rather than [built-in options]


avdgrinten (approved):
LGTM.

PR #150, opened 2021-12-04
add a faster make-image.py

ArsenArsen:


scripts/init-sysroot.sh
```diff
@@ -0,0 +1,63 @@
...
+
+amd64_init() {
+	install -dm 750 "${sysroot}/home/managarm"
+	install -d "${sysroot}/boot/managarm"
+	install -d "${sysroot}/boot/grub"
+
+	# install kernel, prekernel
+	_umb="${sysroot}/usr/managarm/bin"
+	install_s -s "${_umb}"/eir-stivale "${sysroot}/boot/managarm"
+	install_s -s "${_umb}"/eir-mb1 "${sysroot}/boot/managarm"
```
Dennisbonke, line 32: We should really add `eir-mb2` here too. We build it but it never gets installed. While we're at it, we should update the grub entries to add multiboot 2 options too.

scripts/update-image.py
```diff
@@ -1,599 +1,1031 @@
...
+                "-S",
+                "512",
+                "--offset",
+                str(start),
+                self.image,
+                str(size // 1024),
+            ]
+            logged_check_call(cmd)
+
+            # manupilate it using mtools
```
Dennisbonke, line 821: Typo, `manupilate` should be `manipulate`.

Dennisbonke (changes requested):
Looks ok to me, but I'll let @avdgrinten take a look too. One small nitpick tho.

scripts/.flake8
```diff
@@ -0,0 +1,2 @@
+[flake8]
+max-line-length = 88
```
avdgrinten, line 2: Why 88 and not 100?

ArsenArsen: Matches `black`.

avdgrinten: In xbstrap, we configure black to use a line length of 99, we should probably also do that here.

ArsenArsen: ah, I didn't remember doing that, will change this evening

avdgrinten (approved):

PR #151, opened 2021-12-21
app-emulation: add qemu port

qookei:
Some notes:
 - The second patch should be removed once managarm/mlibc#311 is merged.
 - Running qemu in a vm with more than one core may cause a kernel page fault.

### DIFF
--- a/bootstrap.d/app-emulation.yml
+++ b/bootstrap.d/app-emulation.yml
@@ -1,0 +1,43 @@
+packages:
+  - name: qemu
+    architecture: '@OPTION:arch@'
+    metadata:
+      summary: A generic and open source machine emulator and virtualizer
+      description: A generic and open source machine emulator and virtualizer
+      spdx: 'GPL-2.0'
+      website: 'https://qemu.org/'
+      maintainer: 'Kacper Słomiński <qookie@managarm.org>'
+      categories: ['app-emulation']
+    source:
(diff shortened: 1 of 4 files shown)


bootstrap.d/app-emulation.yml
```diff
@@ -0,0 +1,43 @@
...
+      - sdl2
+      - pcre
+      - libiconv
+    configure:
+      - args:
+        - '@THIS_SOURCE_DIR@/configure'
+        - '--cross-prefix=@OPTION:arch-triple@-'
+        - '--host=@OPTION:arch-triple@'
+        - '--prefix=/usr'
+        - '--target-list=i386-softmmu,x86_64-softmmu,aarch64-softmmu'
```
Dennisbonke, line 33: Maybe add the riscv target too for the fun, considering we're working on a riscv port too?

Dennisbonke (commented):
One small comment, and this isn't blocked on mlibc at all at the moment right?

qookei:
Nope it's not blocked on mlibc, it should compile just fine with current master.

bootstrap.d/app-emulation.yml
qookei: Good point, enabled both riscv32 and riscv64 softmmu targets.

Dennisbonke:
LGTM

PR #148, opened 2021-11-28
Various updates

Dennisbonke:
This PR updates the following packages:

- `mesa`, update from version 21.1.4 to 21.2.5;
- `cmake`, update from version 3.20.5 to 3.22.0;
- `curl`, update from version 7.77.0 to 7.80.0;
- `freetype`, update from version 2.10.2 to 2.11.0;
- `nano`, update from version 5.8 to 5.9;
- `libuv`, update from version 1.41.1 to 1.42.0;
- `libxfont2`, update from version 2.0.4 to 2.0.5;
- `libxft`, update from version 2.3.3 to 2.3.4;
- `libjpeg-turbo`, update from version 2.1.0 to 2.1.2;
- `pango`, update from version 1.48.1 to 1.48.10;
- `libepoxy`, update from version 1.5.8 to 1.5.9;
- `harfbuzz`, update from version 2.8.2 to 3.1.2;
- `glib`, update from version 2.68.3 to 2.70.1;
- `libarchive`, update from version 3.5.1 to 3.5.2;
- `libgpg-error`, update from version 1.42 to 1.43;
- `libgcrypt`, update from version 1.8.5 to 1.9.4;
- `libsoup`, update from version 2.72.0 to 2.74.2;
- `wget`, update from version 1.21.1 to 1.21.2;
- `libxkbcommon`, update from version 1.3.0 to 1.3.1;
- `wayland-protocols`, update from version 1.21 to 1.24;
- `libinput`, update from version 1.16.4 to 1.19.2;
- `xorg-proto`, update from version 2021.4 to 2021.5;
- `libxi`, update from version 1.7.10 to 1.8;
- `glib-network`, update from version 2.68.2 to 2.70.0;
- `icu`, update from version 69.1 to 70.1;
- `libevdev`, update from version 1.9.1 to 1.12.0;
- `libdrm`, update from version 2.4.107 to 2.4.109;
- `fribidi`, update from version 1.0.10 to 1.0.11;
- `xkeyboard-config`, update from version 2.33 to 2.34;
- `sqlite`, update from version 3.36.0 to 3.37.0;
- `diffutils`, update from version 3.7 to 3.8;
- `grep`, update from version 3.5 to 3.7;
- `libffi`, update from version 3.3 to 3.4.2;
- `bc`, update from version 3.2.4 to 5.2.1;
- `openssl`, update from version 1.1.1k to 1.1.1l;
- `gdbm`, update from version 1.20 to 1.22;
[...]

PR #153, opened 2022-01-14
weston: bump to 10.0.0

no92:
This is a tracking PR for updating weston to an up-to-date version.

This PR depends on:
- [x] bringing back and testing Xwayland
- [x] merging managarm/managarm#406
- [x] merging #148, more specifically 11d8b8e
- [x] merging managarm/mlibc#355
- [x] check why `sh` is used in weston-terminal instead of `bash` as before
- [x] fix missing fonts in window titlebars
- [x] update to weston 10 upon release

There might be some more changes needed, I will keep this list updated as we move forward.


no92:
Updated to weston 10 and cherry-picked the `wayland-protocols` commit this depends on. Ready for review and testing.

The `memfd_create` changes discussed on Discord are not of concern here, as it is only an optional feature that is disabled if `memfd_create` is not present in headers, as it currently is on mlibc master. Stubbing that without implementing it would, however, break weston.

no92:
Dropped the updates to `libdrm` and `wayland-protocols`, as #148 landed them already.

patches/weston/0002-Add-xwayland-support-for-managarm.patch
```diff
@@ -1,30 +0,0 @@
-From 717a54091fe9e2b71d5e00c44b776d8c2c33b0b1 Mon Sep 17 00:00:00 2001
```
Dennisbonke, line 1: Did you test X based apps without this patch? I'm pretty sure it remains needed and I'm going to double check that myself before we accidentally break Xwayland.

Dennisbonke (changes requested):
No changes needed at the moment, but holding until managarm/mlibc#370 and managarm counterpart are merged and I got a chance at testing Xwayland being operational considering you removed the patch for that.

patches/weston/0002-Add-xwayland-support-for-managarm.patch
no92: Xwayland seems to work for me. I tested `xclock` and some command-line utilities (I don't remember right now) that operated correctly.

avdgrinten:
Can this be merged now? Since we got the clarification that Xwayland works.

Dennisbonke:
> Can this be merged now? Since we got the clarification that Xwayland works.

I plan on testing this today and merging if it works (which seems to be the case).

Dennisbonke (changes requested):
Holding as during testing it consistently crashes for me with memfd. Also holding on the fact that other packages break with memfd (mesa is one notable contender).

no92:
Dennis' crashes were traced to memfd not setting `_link`, which seems to cause nullptr dereferences in fstatat handling. Apparently, the flag `AT_EMPTY_PATH` is also set in those requests.

Dennisbonke:
On that note, #160 seems to unbreak `mesa`, but without a proper testbed I can't confirm that.

Dennisbonke:
Holding until managarm/managarm#418, managarm/mlibc#370 and #160 are merged

Dennisbonke (approved):
LGTM now

PR #155, opened 2022-01-15
X11 libraries fixes

Dennisbonke:
The Vinix team found several issues with our X libraries involving accidental linkage to libraries from the host. They fixed that problem for themselves and alerted us to it, I've backported their fix so that we're no longer affected by it as well.


avdgrinten:
Does this affect all the packages in the diff? Or are the changes applied just in case?

Dennisbonke:
> Does this affect all the packages in the diff? Or are the changes applied just in case?

Applied just in case

PR #159, opened 2022-02-06
mesa: add patch for memfd

no92:

PR #160, opened 2022-02-06
Update mesa to 21.3.5 and use libglvnd

Dennisbonke:
This PR updates `mesa` to version 21.3.5 and makes `mesa` use `libglvnd`. While not a hard dependency to use this at the moment, if we ever want to do `glamor` + `egl` + `wayland` things in `Xorg` (not `Xwayland`) we'll need to use this. Also a very initial and not really noteworthy step to nvidia drivers if that ever happens I guess.

Fixes what #159 was doing but properly.
Unblocks managarm/managarm#418

### DIFF
--- a/bootstrap.d/media-libs.yml
+++ b/bootstrap.d/media-libs.yml
@@ -502,8 +502,8 @@ packages:
     source:
       subdir: 'ports'
       git: 'https://gitlab.freedesktop.org/mesa/mesa.git'
-      tag: 'mesa-21.2.5'
-      version: '21.2.5'
+      tag: 'mesa-21.3.5'
+      version: '21.3.5'
       tools_required:
         - host-pkg-config
     tools_required:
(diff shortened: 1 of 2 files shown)


avdgrinten (approved):
LGTM. Why `-Dasm=disabled`?

Why is this a draft PR?

Dennisbonke:
> LGTM. Why `-Dasm=disabled`?
> 
> 
> 
> Why is this a draft PR?

Uhh I never tested with `-Dasm=enabled`. I'll test that and if it works update the PR and mark it as ready.

Dennisbonke:
> > LGTM. Why `-Dasm=disabled`?
> > Why is this a draft PR?
> 
> Uhh I never tested with `-Dasm=enabled`. I'll test that and if it works update the PR and mark it as ready.

After researching it the only use of assembly that I could find is when certain intrinsics aren't available. `libglvnd` also gives an error because Managarm isn't known. I'll leave it disabled.

Dennisbonke:
Holding until managarm/managarm#418 and managarm/mlibc#370 are merged.

PR #165, opened 2022-02-14
Upgrade LLVM to version 13.0.1

Dennisbonke:
This PR upgrades LLVM to version 13.0.1, it seems like upstream finally fixed our codegen issues and everything builds fine for me. I did have to do a few small changes to the patches, so someone should double check them before merging.


patches/llvm/0002-Add-managarm-target.patch
```diff
@@ -285,13 +288,13 @@ index 000000000..3639920fd
...
 +
-+  const std::string OSLibDir = getOSLibDir(Triple, Args);
++  const std::string OSLibDir = std::string(getOSLibDir(Triple, Args));
 +  const std::string MultiarchTriple = getMultiarchTriple(D, Triple, SysRoot);
 +
 +  // Add the multilib suffixed paths where they are available.
 +  if (GCCInstallation.isValid()) {
 +    const llvm::Triple &GCCTriple = GCCInstallation.getTriple();
-+    const std::string &LibPath = GCCInstallation.getParentLibPath();
++    const std::string &LibPath = std::string(GCCInstallation.getParentLibPath());
```
avdgrinten, line 297: Are these `std::string` casts also done in toolchains other than Managarm?

patches/llvm/0002-Add-managarm-target.patch
```diff
@@ -533,13 +536,18 @@ index 000000000..3639920fd
...
++                                               TripleStr))
 +    return;
 +
++  // The primary search for libstdc++ supports multiarch variants.
++  // if (addLibStdCXXIncludePaths(LibDir.str() + "/../include",
++  //                              "/c++/" + Version.Text, TripleStr,
++  //                              GCCMultiarchTriple, TargetMultiarchTriple,
++  //                              Multilib.includeSuffix(), DriverArgs, CC1Args))
++  //   return;
++
```
avdgrinten, line 550: Where does this come from?

patches/llvm/0002-Add-managarm-target.patch
```diff
@@ -557,9 +565,9 @@ index 000000000..3639920fd
 +  };
 +
 +  for (const auto &IncludePath : LibStdCXXIncludePathCandidates) {
-+    if (addLibStdCXXIncludePaths(IncludePath, /*Suffix*/ "", TripleStr,
-+                                 /*GCCMultiarchTriple*/ "",
-+                                 /*TargetMultiarchTriple*/ "",
++    if (addLibStdCXXIncludePaths(IncludePath, /*Suffix*/ TripleStr,
++                                 /*GCCMultiarchTriple*/
++                                 /*TargetMultiarchTriple*/
```
avdgrinten, line 570: The comments now look very strange here.

patches/llvm/0002-Add-managarm-target.patch
```diff
@@ -68,6 +69,8 @@ index 70fac030b..f1f0c30c3 100644
 +      Builder.defineMacro("_REENTRANT");
 +    if (Opts.CPlusPlus)
 +      Builder.defineMacro("_GNU_SOURCE");
++    if (this->HasFloat128)
++      Builder.defineMacro("__FLOAT128__");
```
avdgrinten, line 73: Why is this done?

patches/llvm/0002-Add-managarm-target.patch
Dennisbonke: Yep, verified it with the Linux toolchain and this is what they did. Without the casts it doesn't compile so I presume LLVM changed some internal thing to require this.

patches/llvm/0002-Add-managarm-target.patch
Dennisbonke: The Generic GCC thing is from the Linux toolchain, and I couldn't find a matching call for the other if statement anywhere. As it doesn't compile and my initial fix is what is done a bit further down in the patch already, I figured we could probably remove it.

patches/llvm/0002-Add-managarm-target.patch
Dennisbonke: You're right, I'll remove them.

patches/llvm/0002-Add-managarm-target.patch
Dennisbonke: The Linux toolchain does this too, and we set `HasFloat128` to true for some targets, I figured we should probably add this then.

patches/llvm/0002-Add-managarm-target.patch
avdgrinten: So the "primary search for libstdc++" comes from our old patch and has no equivalent in the Linux toolchain anymore? If yes, we can probably delete it.

Dennisbonke: Seems like it. I'll delete it.

no92:
Apart from the cleanup, this doesn't seem to break weston, kmscon or a few handful of applications I tested. However, the mmap tests of `posix-tests` fail for me, but I'm not sure whether that's due to LLVM 13 to begin with (vs. for example, some local change I have applied).

Overall, I guess I give this a cautious thumbs-up then. Having CI build this and everyone being able to pull feels worthwhile, given that building this PR took me over 4.5 hours.

no92:
Turns out the mmap failures were due to local changes, so this PR gets a :+1: from me.

no92 (approved):

64:
`exa -l` breakage seems unrelated (rtdl issue)

PR #167, opened 2022-02-16
rust: update rustc to 1.58.1

64:
Threads are still broken as per https://github.com/managarm/mlibc/issues/389.


avdgrinten:
Are threads also broken in 1.57?

Dennisbonke (changes requested):
Looks good to me but we can merge some patches back into one I think. Personally, I'd merge `rust-libc` patch 3 and 4 into 1, and `rust` patch 3 into patch 1.

Dennisbonke:
> Are threads also broken in 1.57?

Seems like they are.

64:
> Are threads also broken in 1.57?

Yes, it's purely a mlibc issue as far as we can tell

avdgrinten:
Alright, then that doesn't block merging.

64:
Confirmed that with managarm/mlibc#392 applied, threads work again:

image

PR #168, opened 2022-02-17
Add various missing post install hooks

Dennisbonke:
This PR adds several missing post install hooks to glib and related packages. Fixing this allows for a better experience in Managarm.

### DIFF
--- a/bootstrap.d/dev-libs.yml
+++ b/bootstrap.d/dev-libs.yml
@@ -223,7 +223,7 @@ packages:
       - zlib
       - libiconv
       - libintl
-    revision: 2
+    revision: 3
     configure:
       - args:
         - 'meson'
@@ -242,6 +242,10 @@ packages:
       - args: ['ninja', 'install']
         environ:
           DESTDIR: '@THIS_COLLECT_DIR@'
+    scripts:
+        post_install:
+          - args: ['glib-compile-schemas /usr/share/glib-2.0/schemas']
+          - args: ['gio-querymodules /usr/lib/gio/modules']
 
   - name : gmp
     labels: [aarch64]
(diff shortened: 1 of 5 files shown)


avdgrinten:
Do all these operations really take a few seconds? If not, I'd just delete the sentence (we don't need it at every hook, we just have it in the fc-cache invocation since that can truly take a long time.

bootstrap.d/dev-libs.yml
```diff
@@ -242,6 +242,10 @@ packages:
       - args: ['ninja', 'install']
         environ:
           DESTDIR: '@THIS_COLLECT_DIR@'
+    scripts:
+        post_install:
+          - args: 'glib-compile-schemas /usr/share/glib-2.0/schemas'
+          - args: 'gio-querymodules /usr/lib/gio/modules'
```
avdgrinten, line 248: Why is this a shell invocation and not just a argument array (same for the other instances)? AFAICS the previous patch didn't work because it was missing a comma in the arg array. But `['glib-compile-schemas', '/usr/share/glib-2.0/schemas']` should work, right?

Dennisbonke: You might be on to something there. I’ll change it around and test it. If that works, I’ll use that over shell invocations.

Dennisbonke:
This still doesn't work properly unfortunately. Marking as draft pending further investigation.

Dennisbonke:
Some of these are shell invocations, after a long battle that is how I got it to work. At this point in time all hooks _should_ work, so this can be merged now.

Dennisbonke:
At the request of @ArsenArsen, I've reworded several commits so they reflect the changes more accurately. It _should_ be good to go now, but if @avdgrinten and @ArsenArsen want to take another look before hitting that merge button that would be great.

avdgrinten (approved):

PR #171, opened 2022-02-22
havoc: Add port

Dennisbonke:
This PR adds a port of `havoc`, a minimalistic terminal emulator for weston.

Blocked on managarm/mlibc#398

### DIFF
--- a/bootstrap.d/gui-apps.yml
+++ b/bootstrap.d/gui-apps.yml
@@ -1,4 +1,54 @@
 packages:
+  - name: havoc
+    architecture: '@OPTION:arch@'
+    metadata:
+      summary: Minimal terminal emulator for wayland 
+      description: This package contains a minimalistic terminal emulator for wayland.
+      spdx: 'MIT'
+      website: 'https://github.com/ii8/havoc'
+      maintainer: "Dennis Bonke <dennis@managarm.org>"
+      categories: ['gui-apps']
+    source:
+      subdir: 'ports'
+      git: 'https://github.com/ii8/havoc.git'
+      tag: '0.4.0'
+      version: '0.4.0'
+    tools_required:
+      - system-gcc
+      - host-pkg-config
+      - virtual: pkgconfig-for-target
+        triple: "@OPTION:arch-triple@"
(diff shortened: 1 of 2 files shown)


bootstrap.d/gui-apps.yml
```diff
@@ -0,0 +1,50 @@
...
+          PKG_CONFIG_SYSROOT_DIR: '@BUILD_ROOT@/system-root'
+          PKG_CONFIG_LIBDIR: '@BUILD_ROOT@/system-root/usr/lib/pkgconfig:@BUILD_ROOT@/system-root/usr/share/pkgconfig'
+          PREFIX: '/usr'
+          DESTDIR: '@THIS_COLLECT_DIR@'
+          CC: "@OPTION:arch-triple@-gcc"
+          CXX: "@OPTION:arch-triple@-g++"
+          CPP: "@OPTION:arch-triple@-cpp"
+      - args: ['mkdir', '-p', '@THIS_COLLECT_DIR@/home/managarm/.config']
+      - args: ['mkdir', '-p', '@THIS_COLLECT_DIR@/root/.config']
+      - args: ['cp', '@THIS_SOURCE_DIR@/havoc.cfg', '@THIS_COLLECT_DIR@/home/managarm/.config/havoc.cfg']
```
no92, line 49: Any specific reason why this needs to be done in both `/home/managarm` and `/root`?

no92 (approved):
Locally tested, works good. On sway, there is some space in between the terminal border and window decoration, which might however be due to sway/wlroots.

bootstrap.d/gui-apps.yml
Dennisbonke: It's per user, so to future proof it we also add the file for the managarm user.

avdgrinten (approved):
LGTM.

bootstrap.d/gui-apps.yml
ArsenArsen: didn't we decide not to install files in /home (since we shouldn't be hard coding any user files)?

PR #172, opened 2022-02-22
Add several missing X.Org libraries

Dennisbonke:
This PR adds several X.Org libraries listed in the BLFS page for X libraries.

### DIFF
--- a/bootstrap.d/x11-libs.yml
+++ b/bootstrap.d/x11-libs.yml
@@ -560,6 +560,53 @@ packages:
         environ:
           DESTDIR: '@THIS_COLLECT_DIR@'
 
+  - name: libxcomposite
+    architecture: '@OPTION:arch@'
+    metadata:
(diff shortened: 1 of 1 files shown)


avdgrinten:
What are these libraries used for? We should not add libraries without consumers. Otherwise, we don't really know whether they work or not.

Dennisbonke:
> What are these libraries used for? We should not add libraries without consumers. Otherwise, we don't really know whether they work or not.

Libxinerama and Libxcomposite are quite commonly used, and I plan on using them for the X backend in webkitgtk, among other things. Libxscrnsaver is for screen savers, libxres seems to be listed as a dependency for some X apps and xorg-server in gentoo, and libdmx is also used by Xorg-server. I haven't updated the consumers to use them yet so at this moment in time they are unused in master.

avdgrinten (approved):
LGTM. Let's merge this as a preparation for WebKit.

PR #173, opened 2022-03-03
scripts: Make the qemu title bar reflect which Managarm we're running

Dennisbonke:
A small change that looks nice and doesn't interfere with anything else, should be good to merge.

### DIFF
--- a/scripts/vm-util.py
+++ b/scripts/vm-util.py
@@ -32,6 +32,8 @@ def do_qemu(args):
 		'-m', '2048',
 	]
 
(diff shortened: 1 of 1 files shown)


ArsenArsen (changes requested):
making this something like `f"managarm {args.arch}` is good enough and isn't as error prone

Dennisbonke:
> making this something like `f"managarm {args.arch}"` is good enough and isn't as error prone

As requested, changed to that. Still works so should be good now.

ArsenArsen (approved):

PR #174, opened 2022-04-03
llvm: Update from version 13.0.1 to 14.0.1

Dennisbonke:
LLVM 14 has dropped, and that can mean only one thing, toolchain update time!

### DIFF
--- a/bootstrap.yml
+++ b/bootstrap.yml
@@ -106,8 +106,8 @@ sources:
   - name: llvm
     subdir: 'ports'
     git: 'https://github.com/llvm/llvm-project'
-    tag: 'llvmorg-13.0.1'
-    version: '13.0.1'
+    tag: 'llvmorg-14.0.1'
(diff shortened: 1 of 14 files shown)


bootstrap.yml
```diff
@@ -969,6 +969,8 @@ packages:
         - '-DLLVM_HOST_TRIPLE=@OPTION:arch-triple@'
         # Disable linking against ncurses, which we do not build with -fPIC.
         - '-DLLVM_ENABLE_TERMINFO=OFF'
+        # Suppress developer warnings
+        - '-Wno-dev'
```
avdgrinten, line 973: Why do we do that?

Dennisbonke: Because it spams the console during LLVM configuration and we aren’t LLVM developers.

64 (approved):
LGTM

PR #169, opened 2022-02-19
Add a port of GTK3 and related dependencies

Dennisbonke:
This PR splits off the `gtk3` part from #152. Several issues still remain, including `dbus` not working properly and `gtk3` missing `pwrite()`, potentially among other things.

This PR adds the following ports:
- `libxslt`,
- `dbus`,
- `at-spi2-core`,
- `at-spi2-atk`,
- `gtk+-3`.

Blocked on #168 and a draft pending the things I mentioned above.
Blocked on #176

PR #152, opened 2022-01-02
Port webkitgtk

Dennisbonke:
This PR adds a port of `webkitgtk`. This PR remains a draft for the foreseeable future and only serves to help others debug various issues in `webkitgtk`.

This PR adds the following ports:
- `libwpe` (needed for newer versions);
- `wpebackend-fdo` (needed for newer versions);
- `webkitgtk`.

Blocked on #169 
Blocked on #176 
Blocked on managarm/managarm#525
Blocked on managarm/managarm#526

PR #176, opened 2022-04-19
Add several libraries to be used by Qt and GTK

Dennisbonke:
This PR splits off the `libwebp` port from #169, the `gstreamer` and `gst-plugins-base` ports from #152 and adds a port of `libtiff`, in preparation of `Qt6` and `GTK` to use them.

Blocked on managarm/mlibc#406


qookei (approved):
LGTM

PR #178, opened 2022-04-27
Add several dependencies for QtWebkit

Dennisbonke:
This PR aims to add all the needed dependencies for QtWebkit to compile. As this is still in progress, this PR is a draft.

- `pciutils`.

### DIFF
--- a/patches/pciutils/0001-Add-initial-Managarm-support.patch
+++ b/patches/pciutils/0001-Add-initial-Managarm-support.patch
@@ -1,0 +1,147 @@
+From f6c4152a19b2430b491ff9b0c0e16a2b368ca60d Mon Sep 17 00:00:00 2001
+From: Dennis Bonke <admin@dennisbonke.com>
+Date: Wed, 27 Apr 2022 23:11:14 +0200
+Subject: [PATCH] Add initial Managarm support
+
+Signed-off-by: Dennis Bonke <admin@dennisbonke.com>
+---
+ lib/filter.c        |  8 ++++----
+ lib/i386-io-linux.h | 25 +++++++++++++++++++++++++
+ lib/i386-ports.c    | 20 ++++++++++++++++++++
+ lib/sysdep.h        |  6 +++---
+ 4 files changed, 52 insertions(+), 7 deletions(-)
+
+diff --git a/lib/filter.c b/lib/filter.c
+index b881b6b..294c69b 100644
+--- a/lib/filter.c
++++ b/lib/filter.c
+@@ -304,20 +304,20 @@ pci_filter_match_v30(struct pci_filter_v30 *f, struct pci_dev *d)
+ 
+ STATIC_ALIAS(void pci_filter_init(struct pci_access *a, struct pci_filter *f), pci_filter_init_v38(a, f));
+ SYMBOL_VERSION(pci_filter_init_v30, pci_filter_init@LIBPCI_3.0);
+-SYMBOL_VERSION(pci_filter_init_v38, pci_filter_init@LIBPCI_3.3);
++//SYMBOL_VERSION(pci_filter_init_v38, pci_filter_init@LIBPCI_3.3);
+ SYMBOL_VERSION(pci_filter_init_v38, pci_filter_init@@LIBPCI_3.8);
+ 
+ STATIC_ALIAS(char *pci_filter_parse_slot(struct pci_filter *f, char *str), pci_filter_parse_slot_v38(f, str));
+ SYMBOL_VERSION(pci_filter_parse_slot_v30, pci_filter_parse_slot@LIBPCI_3.0);
+-SYMBOL_VERSION(pci_filter_parse_slot_v38, pci_filter_parse_slot@LIBPCI_3.3);
++//SYMBOL_VERSION(pci_filter_parse_slot_v38, pci_filter_parse_slot@LIBPCI_3.3);
+ SYMBOL_VERSION(pci_filter_parse_slot_v38, pci_filter_parse_slot@@LIBPCI_3.8);
+ 
+ STATIC_ALIAS(char *pci_filter_parse_id(struct pci_filter *f, char *str), pci_filter_parse_id_v38(f, str));
+ SYMBOL_VERSION(pci_filter_parse_id_v30, pci_filter_parse_id@LIBPCI_3.0);
+-SYMBOL_VERSION(pci_filter_parse_id_v38, pci_filter_parse_id@LIBPCI_3.3);
++//SYMBOL_VERSION(pci_filter_parse_id_v38, pci_filter_parse_id@LIBPCI_3.3);
+ SYMBOL_VERSION(pci_filter_parse_id_v38, pci_filter_parse_id@@LIBPCI_3.8);
+ 
+ STATIC_ALIAS(int pci_filter_match(struct pci_filter *f, struct pci_dev *d), pci_filter_match_v38(f, d));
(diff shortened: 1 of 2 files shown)


patches/ninja/0001-Add-support-for-NINJAJOBS.patch
```diff
@@ -0,0 +1,29 @@
...
+--- a/src/ninja.cc
++++ b/src/ninja.cc
+@@ -227,6 +227,11 @@ void Usage(const BuildConfig& config) {
+ 
+ /// Choose a default value for the -j (parallelism) flag.
+ int GuessParallelism() {
++  int   j = 0;
++  char* jobs = getenv( "NINJAJOBS" );
++  if ( jobs != NULL ) j = atoi( jobs );
++  if ( j > 0 ) return j;
```
avdgrinten, line 22: Why do we patch ninja instead of just using the `-j` flag? Patching ninja doesn't sound really appealing to me.

patches/ffmpeg/0001-Add-initial-Managarm-support.patch
```diff
@@ -0,0 +1,26 @@
...
+diff --git a/libavutil/cpu.c b/libavutil/cpu.c
+index 6bd0f07..4d692e4 100644
+--- a/libavutil/cpu.c
++++ b/libavutil/cpu.c
+@@ -280,7 +280,7 @@ int av_cpu_count(void)
+ #if HAVE_WINRT
+     SYSTEM_INFO sysinfo;
+ #endif
+-#if HAVE_SCHED_GETAFFINITY && defined(CPU_COUNT)
++#if HAVE_SCHED_GETAFFINITY && defined(CPU_COUNT) && !defined(__managarm__)
```
avdgrinten, line 20: Why is this done?

patches/pciutils/0001-Add-initial-Managarm-support.patch
```diff
@@ -0,0 +1,147 @@
...
+diff --git a/lib/filter.c b/lib/filter.c
+index b881b6b..294c69b 100644
+--- a/lib/filter.c
++++ b/lib/filter.c
+@@ -304,20 +304,20 @@ pci_filter_match_v30(struct pci_filter_v30 *f, struct pci_dev *d)
+ 
+ STATIC_ALIAS(void pci_filter_init(struct pci_access *a, struct pci_filter *f), pci_filter_init_v38(a, f));
+ SYMBOL_VERSION(pci_filter_init_v30, pci_filter_init@LIBPCI_3.0);
+-SYMBOL_VERSION(pci_filter_init_v38, pci_filter_init@LIBPCI_3.3);
++//SYMBOL_VERSION(pci_filter_init_v38, pci_filter_init@LIBPCI_3.3);
```
avdgrinten, line 23: Why is this done?

patches/pciutils/0001-Add-initial-Managarm-support.patch
```diff
@@ -0,0 +1,147 @@
...
++}
++
++static unsigned short int inw(unsigned short int port) {
++	return 0;
++}
++
++static unsigned int inl(unsigned short int port) {
++	return 0;
++}
++#endif
```
avdgrinten, line 109: This does not look correct. We should not silently return wrong values (?).

patches/pciutils/0001-Add-initial-Managarm-support.patch
```diff
@@ -0,0 +1,147 @@
...
++#if defined PCI_HAVE_LINUX_BYTEORDER_H && !defined __managarm__
+ 
+ #include <asm/byteorder.h>
+ #define cpu_to_le16 __cpu_to_le16
+@@ -39,7 +39,7 @@ typedef u16 word;
+ 
+ #else
+ 
+-#ifdef PCI_OS_LINUX
++#if defined PCI_OS_LINUX && !defined __managarm__
```
avdgrinten, line 132: Why is this done?

patches/snappy/snappy-1.1.9_gcc_inline.patch
```diff
@@ -0,0 +1,67 @@
+From 0c716d435abe65250100c2caea0e5126ac4e14bd Mon Sep 17 00:00:00 2001
```
avdgrinten, line 1: This file does not match our usual patch naming convention.

patches/snappy/snappy-1.1.9_remove-no-rtti.patch
```diff
@@ -0,0 +1,35 @@
+From 516fdcca6606502e2d562d20c01b225c8d066739 Mon Sep 17 00:00:00 2001
```
avdgrinten, line 1: This file does not match our usual patch naming convention.

patches/ninja/0001-Add-support-for-NINJAJOBS.patch
Dennisbonke: If ninja is invoked by a different build system or a script (for example, qtwebengine, which builds with cmake but then builds chromium with ninja), an user can't pass the `-j` flag. With this patch, we can also use the `NINJAJOBS` environment variable. The patch was sourced from BLFS.

patches/pciutils/0001-Add-initial-Managarm-support.patch
Dennisbonke: I got strange linker errors related to this, so I removed it. If requested, we could investigate it and fix those instead.

patches/pciutils/0001-Add-initial-Managarm-support.patch
Dennisbonke: Indeed we shouldn't, but at the time of creating this patch the libc wrapper functions didn't exist yet and I was too lazy to patch them in properly. With the correct mlibc PR merged, we might be able to drop this part of the patch completely.

patches/snappy/snappy-1.1.9_gcc_inline.patch
Dennisbonke: The patch was taken as-is from gentoo, so I kept the name intact as well.

patches/snappy/snappy-1.1.9_remove-no-rtti.patch
Dennisbonke: The patch was taken as-is from gentoo, so I kept the name intact as well.

patches/pciutils/0001-Add-initial-Managarm-support.patch
Dennisbonke: I honestly don't remember anymore, we might be able to just drop this part.

patches/snappy/snappy-1.1.9_remove-no-rtti.patch
avdgrinten: The issue is that xbstrap sorts patches by their file name, so if we're missing the `0001-` prefix, they'll be applied in "random" orders.

patches/ninja/0001-Add-support-for-NINJAJOBS.patch
qookei: When using the `build` subcommand of cmake, you can pass arguments to make/ninja like so: `cmake build -- -jN`

Dennisbonke: Interesting, I didn't know that. In that case it's possible to drop the patch, tho I don't mind keeping it either.

PR #182, opened 2022-05-20
bootstrap.yml: install cxxshim and frigg headers to /usr/share

64:
As discussed on Discord. This allows cxxshim and frigg to be found via `pkg-config` rather than needing to do `ln` hacks.

Some things may break if they incorrectly use frigg without declaring it as a dependency, see e.g https://github.com/managarm/managarm/pull/453.

### DIFF
--- a/bootstrap.yml
+++ b/bootstrap.yml
@@ -1047,6 +1047,9 @@ packages:
       - host-managarm-tools
       - host-protoc
       - kernel-gcc
+      - host-pkg-config
+      - virtual: pkgconfig-for-target
+        triple: "@OPTION:arch-triple@"
     pkgs_required:
       - frigg
     revision: 2
@@ -1089,6 +1092,8 @@ packages:
       - host-pkg-config
       - host-protoc
       - system-gcc
+      - virtual: pkgconfig-for-target
+        triple: "@OPTION:arch-triple@"
     pkgs_required:
       - mlibc
       - eudev
@@ -1197,6 +1202,8 @@ packages:
       - bootstrap-system-gcc
       - host-protoc
       - host-pkg-config
+      - virtual: pkgconfig-for-target
+        triple: "@OPTION:arch-triple@"
     pkgs_required:
       - mlibc-headers
       - frigg
--- a/bootstrap.yml
+++ b/bootstrap.yml
@@ -143,8 +143,6 @@ sources:
     version: '0.0pl@ROLLING_ID@'
     sources_required:
       - cralgo
-      - cxxshim
-      - frigg
       - lai
       - libarch
       - libasync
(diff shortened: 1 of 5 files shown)


ArsenArsen:
why not /usr/include/$name instead?

64:
> why not /usr/include/$name instead?

cc @avdgrinten

avdgrinten:
That's even more confusing since the path to the headers is then `/usr/include/$name/$name`.

64:
maybe as a compromise, we could install to `/usr/include/freestanding/$name`?

ArsenArsen:
well that'd only be necessary for cxxshim right? Frigg doesn't conflict with anything, and hence could reside globally?

avdgrinten:
In principle yes, but doing it for all freestanding libs has the advantage that we cannot accidentally include non-freestanding headers.

I would still prefer installing to `/usr/share` instead of `/usr/include/freestanding`. We could symlinks `/usr/include/$name` to `/usr/share/$name/include/$name` but that would only be useful for build systems that do not use pkg-config anyway.

avdgrinten:
We should also consider adding `--wrap-mode=nofallback` to the Meson packages that do not use the "symlinked subproject" hack anymore.

bootstrap.yml
```diff
@@ -1133,6 +1135,28 @@ packages:
           'DESTDIR': '@THIS_COLLECT_DIR@'
         quiet: true
 
+  - name: cxxshim
+    labels: [aarch64]
```
Dennisbonke, line 1139: While insignificant, we can probably add risc-v as a label here too.

bootstrap.yml
```diff
@@ -1133,6 +1135,28 @@ packages:
...
+    architecture: '@OPTION:arch@'
```
Dennisbonke, line 1140: Headers only, thus arch independent? Mark as noarch in that case.

bootstrap.yml
```diff
@@ -760,6 +757,8 @@ packages:
         - 'meson'
         - '--prefix=/usr'
         - '--libdir=lib'
+        # Install to /usr/share to avoid conflicts with standard C++ headers.
```
Dennisbonke, line 760: Going to put it here, but the same comments from cxxshim apply for frigg, while we’re working on it, we can fix it, saves us fixing later.

bootstrap.yml
```diff
@@ -1133,6 +1135,28 @@ packages:
           'DESTDIR': '@THIS_COLLECT_DIR@'
         quiet: true
 
+  - name: cxxshim
```
Dennisbonke, line 1138: Missing metadata, please add it. We should add missing metadata whenever we’re updating a package. I guess the maintainer of both frigg and cxxshim would be @avdgrinten?

Dennisbonke (changes requested):
Other than the change @avdgrinten requested, looks good to me. My comments are minor and I can live without them, tho the metadata thing really should be fixed.

64:
Blocker: I've just noticed that `host-managarm-tools` also requires frigg headers, but there's no straightforward way to satisfy that requirement now that we don't symlink frigg like we did before.

avdgrinten:
Can this be merged now?

64:
Should be good to go

avdgrinten:
Thank you!

PR #183, opened 2022-05-26
xclock: add missing dependency on util-linux-libs

no92:
Running `xclock -help` without `util-linux-libs` installed complains about missing `libuuid.so.1`, which is solved this way.

### DIFF
--- a/bootstrap.d/x11-apps.yml
+++ b/bootstrap.d/x11-apps.yml
@@ -77,7 +77,8 @@ packages:
       - libxt
       - libxkbfile
       - libiconv
-    revision: 2
+      - util-linux-libs
+    revision: 3
     configure:
       - args:
         - '@THIS_SOURCE_DIR@/configure'


bootstrap.d/x11-apps.yml
```diff
@@ -77,6 +77,7 @@ packages:
       - libxt
       - libxkbfile
       - libiconv
+      - util-linux-libs
     revision: 2
```
Dennisbonke, line 81: We should bump the revision by one so that a new package is generated.

Dennisbonke (changes requested):
Interesting find, I myself saw nothing suggesting that this dependency was needed but I trust your observation. One small change is needed, other than that this looks fine.

avdgrinten:
Why is this lib not required at build time (and the build failure detected by our build server)?

avdgrinten:
Seems that xclock already transitively depends on util-linux-libs. How were you able to install xclock w/o installing util-linux-libs?

ArsenArsen:
huh, how did this build without the dep?

Dennisbonke:
> Seems that xclock already transitively depends on util-linux-libs. How were you able to install xclock w/o installing util-linux-libs?

Good catch, the X stack pulls in python which depends on util-linux-libs. That python dependency is mostly if not fully build time but we can’t specify that in xbstrap at this point in time.

PR #186, opened 2022-06-06
fish: Add port

64:
This PR adds a port of `fish`.

Currently blocked on https://github.com/managarm/mlibc/issues/467.


bootstrap.d/app-shells.yml
```diff
@@ -48,3 +48,37 @@ packages:
       - args: ['cp', '@SOURCE_ROOT@/extrafiles/profile', '@THIS_COLLECT_DIR@/etc']
       - args: ['cp', '@SOURCE_ROOT@/extrafiles/bash.bashrc', '@THIS_COLLECT_DIR@/etc']
       - args: ['ln', '-sf', 'bash', '@THIS_COLLECT_DIR@/usr/bin/sh']
+
+  - name: fish
+    metadata:
+      summary: Friendly interactive SHell
+      description: Smart and user-friendly portable command line shell.
+      spdx: 'GPL-2'
```
Dennisbonke, line 56: `GPL-2` is not the correct SPDX for `fish`. It should be `GPL-2.0-only`. [This](https://spdx.org/licenses/) is an amazing list for SPDX identifiers.

bootstrap.d/app-shells.yml
```diff
@@ -48,3 +48,37 @@ packages:
...
+    architecture: '@OPTION:arch@'
+    source:
+      subdir: 'ports'
+      git: 'https://github.com/fish-shell/fish-shell.git'
+      tag: '3.4.1'
+      version: '3.4.1'
+    tools_required:
+      - system-gcc
+      - host-cmake
+    pkgs_required:
```
Dennisbonke, line 68: Missing dependency on `mlibc`, I'm pretty sure it depends on the libc.

Dennisbonke: Also seems like one can use `libintl` for i18n support, we can specify the dependency for later use.

Dennisbonke (changes requested):
LGTM apart from a few small comments.

PR #187, opened 2022-06-07
zsh: add port

no92:
This adds a zsh port that seems to work for simple command-line use. Building this requires `yodl` being installed in the rootfs, i.e. chroot into the rootfs and run `apt install yodl`. Including this as a host-tool would introduce 2 host-tools with ridiculous build systems.


no92:
Updated and rebased on current master. `zsh` seems to work fine, so I don't see any blockers here any more.

Dennisbonke (approved):
LGTM

PR #188, opened 2022-06-07
docker: add yodl

no92:
Add `yodl` to the Docker image, as for the first build of `zsh` in #187 not to fail.


Dennisbonke (approved):
LGTM

PR #196, opened 2022-06-15
Add telescope+libretls port

ElectrodeYT:


patches/telescope/0001-Managarm-Patches.patch
```diff
@@ -0,0 +1,73 @@
...
+diff --git a/telescope.c b/telescope.c
+index 0a65f41..1991d1b 100644
+--- a/telescope.c
++++ b/telescope.c
+@@ -64,7 +64,7 @@ int			 operating;
+  * "Safe" (or "sandobox") mode.  If enabled, Telescope shouldn't write
+  * anything to the filesystem or execute external programs.
+  */
+-int			safe_mode;
++int			safe_mode = 0;
```
avdgrinten, line 45: This shouldn't do anything since global vars are initialized to zero anyway.

bootstrap.d/dev-libs.yml
```diff
@@ -794,7 +794,34 @@ packages:
       - args: ['make', 'install']
         environ:
           DESTDIR: '@THIS_COLLECT_DIR@'
-
+  - name: libretls
```
avdgrinten, line 797: This port lacks meta data. Is it necessary to use this package? We already have OpenSSL.

patches/telescope/0001-Managarm-Patches.patch
```diff
@@ -0,0 +1,73 @@
...
+ static struct imsgev	*iev_net;
+ 
+@@ -1047,8 +1047,8 @@ main(int argc, char * const *argv)
+ 
+ 	if (getenv("NO_COLOR") != NULL)
+ 		enable_colors = 0;
+-
+-	while ((ch = getopt_long(argc, argv, opts, longopts, NULL)) != -1) {
++	optind = 0;
++	while ((ch = getopt(argc, argv, opts)) != -1) {
```
avdgrinten, line 56: Why is this done?

patches/telescope/0001-Managarm-Patches.patch
```diff
@@ -0,0 +1,73 @@
...
+ 		case 'C':
+ 			exit(ui_print_colors());
+@@ -1128,8 +1128,9 @@ main(int argc, char * const *argv)
+ 		    "telescope already running?");
+ 	}
+ 
++
+ 	/* Start children. */
+-	if (socketpair(AF_UNIX, SOCK_STREAM, PF_UNSPEC, pipe2net) == -1)
++	if (socketpair(AF_UNIX, SOCK_STREAM, 0, pipe2net) == -1)
```
avdgrinten, line 67: Why is this done? If we're missing `PF_UNSPEC`, adding it to mlibc should be simple.

bootstrap.d/dev-libs.yml
ArsenArsen: libretls is a libtls reimplementation of libtls for openssl

patches/telescope/0001-Managarm-Patches.patch
no92: Working around unimplemented/incorrectly implemented functionality in `getopt_long`. The `optind = 0` can be dropped, though.

patches/telescope/0001-Managarm-Patches.patch
Dennisbonke: We have it, however Managarm code asserts that the protocol (3rd argument) is 0. We should change the Managarm code to allow PF_UNSPEC (which is basically pick what you think is best afaik).

bootstrap.d/dev-libs.yml
Dennisbonke: Aside from missing metadata and architecture labels, ports should have one empty line between them. And yes, we need `libretls` or make `telescope` link against `libressl`, which I'm against. Gentoo has `libretls` in their repo's too, so this is fine.

bootstrap.d/dev-libs.yml
```diff
@@ -794,7 +794,34 @@ packages:
...
+      - args:
+        - '@THIS_SOURCE_DIR@/configure'
+        - '--host=@OPTION:arch-triple@'
+        - '--prefix=/usr'
+        - '--disable-static'
+    build:
+      - args: ['make', '-j@PARALLELISM@']
+      - args: ['make', 'install']
+        environ:
+          DESTDIR: '@THIS_COLLECT_DIR@'
```
Dennisbonke, line 824: Same here, newline after a port recipe.

bootstrap.d/www-client.yml
```diff
@@ -53,3 +53,40 @@ packages:
...
+      - system-gcc
+    pkgs_required:
+      - mlibc
+      - libretls
+      - openssl
+      - ncurses
+    configure:
+      - args: 
+        - '@THIS_SOURCE_DIR@/configure'
+        - '--host=x86_64-managarm'
```
Dennisbonke, line 87: We should use `- '--host=@OPTION:arch-triple@'` instead of hardcoding architectures here.

bootstrap.d/www-client.yml
```diff
@@ -53,3 +53,40 @@ packages:
       - args: ['make', 'DESTDIR=@THIS_COLLECT_DIR@', 'install']
         environ:
           DESTDIR: '@THIS_COLLECT_DIR@'
+  - name: telescope
```
Dennisbonke, line 56: Same here, a newline should be added between ports.

Dennisbonke (changes requested):
For a first attempt this is quite decent, I've added my comments that others hadn't made yet.

patches/telescope/0001-Managarm-Patches.patch
no92: See managarm/mlibc#438 for the fix.

ElectrodeYT:
Gonna hold of on this until the various mlibc things are merged

PR #197, opened 2022-06-21
rust: bump cargo version to 1.60.0

64:
Bump cargo version to match rustc. This should fix alacritty on CI.

### DIFF
--- a/bootstrap.d/dev-lang.yml
+++ b/bootstrap.d/dev-lang.yml
@@ -26,10 +26,10 @@ sources:
 
   - name: host-bootstrap-cargo
     subdir: 'ports'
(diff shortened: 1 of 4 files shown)


bootstrap.d/dev-lang.yml
```diff
@@ -87,8 +87,9 @@ tools:
     source:
       subdir: 'ports'
       git: 'https://github.com/rust-lang/cargo.git'
-      tag: '0.53.0'
-      version: '0.53.0'
+      tag: '0.60.0'
+      version: '0.60.0'
+    revision: 2
```
avdgrinten, line 92: Revisions reset when the version increases, so we do not need to bump the revision here.

Dennisbonke (approved):
LGTM

PR #198, opened 2022-07-09
Drop stivale2 boot protocol in favour of multiboot2

mintsuki:
The stivale boot protocols are deprecated and due to be removed in Limine 4.0.

We could move to the Limine boot protocol but I don't think there is a need to do so due to eir already supporting multiboot2 just fine (which is not deprecated).


Dennisbonke (approved):
LGTM

64 (approved):

PR #201, opened 2022-08-07
Small fixes to json-c, zip, unzip and fontconfig

no92:


Dennisbonke (approved):
LGTM

patches/json-c/0001-Add-managarm-specific-changes.patch
```diff
@@ -0,0 +1,25 @@
...
+diff --git a/json_tokener.c b/json_tokener.c
+index 0c09b66..d74153a 100644
+--- a/json_tokener.c
++++ b/json_tokener.c
+@@ -327,7 +327,7 @@ struct json_object *json_tokener_parse_ex(struct json_tokener *tok, const char *
+ 		if (newloc == NULL)
+ 		{
+ 			freelocale(duploc);
+-			return NULL;
++			// return NULL;
```
avdgrinten, line 19: Why is this done?

no92: Due to our stubbing of locale functions, this prevents json-c from parsing input properly.

avdgrinten: Is the resulting code path correct? I'd expect that there is some reason that they return and don't ignore the error here. IMHO such cases should receive comments (or this should be fixed in mlibc).

no92: The parsing after this is correct. The problem that this is running into is that mlibc stubs locale functions, and returns nullptrs. This mostly works fine, but json-c treats this as an error condition and stops parsing, which is not what we want here.

PR #206, opened 2022-08-25
Another batch of updates

Dennisbonke:
The following ports received updates:

- `atk`, from version 2.36.0 to 2.38.0;
- `bash`, from version 5.1.12 to 5.1.16;
- `nspr`, from version 4.33 to 4.34.1;
- `nss`, from version 3.77 to 3.82;
- `ca-certificates`, from version 20220427.3.77 to 20220825.3.82;
- `cairo`, from version 1.17.4 to 1.17.6;
- `cmake`, from version 3.22 to 3.24.1;
- `cups`, from version 2.3.3 to 2.4.2;
- `curl`, from version 7.80.0 to 7.84.0;
- `double-conversion`, from version 3.1.5 to 3.2.1;
- `fontconfig`, from version 2.13.92 to 2.14.0;
- `freetype`, from version 2.11.0 to 2.12.1;
- `fribidi`, from version 1.0.11 to 1.0.12;
- `glib`, from version 2.70.3 to 2.72.3;
- `glib-networking`, from version 2.70.0 to 2.72.2;
- `gnutls`, from version 3.7.2 to 3.7.7;
- `gsettings-desktop-schemas`, from version 3.38.0 to 42.0;
- `openssl`, from version 1.1.1l to 1.1.1q;
- `libarchive`, from version 3.5.2 to 3.6.1;
- `libdrm`, from version 2.4.109 to 2.4.112;
- `libepoxy`, from version 1.5.9 to 1.5.10;
- `libevdev`, from version 1.12.0 to 1.13.0;
- `libgcrypt`, from version 1.9.4 to 1.10.1;
- `libgpg-error`, from version 1.43 to 1.45.

Furthermore, mpfr got his documentation installation directory fixed to use the correct version, thanks @ArsenArsen for pointing that out to me.

This is not everything but I'm breaking it up into smaller parts due to vacation.

PR #207, opened 2022-08-29
Another batch of updates

Dennisbonke:

- `libiconv`, from version 1.16 to 1.17;
- `libjpeg-turbo`, from version 2.1.2 to 2.1.4;
- `libpipeline`, from version 1.5.4 to 1.5.6;
- `libtasn`, from version 4.17.0 to 4.18.0;
- `libtiff`, from version 4.3.0 to 4.4.0;
- `libunistring`, from version 0.9.10 to 1.0;
- `libwebp`, from version 1.2.2 to 1.2.4;
- `libX11`, from version 1.7.3 to 1.8.1;
- `libxcb`, from version 1.14 to 1.15;
- `libxcrypt`, from version 4.4.26 to 4.4.28;
- `libXcursor`, from version 1.2.0 to 1.2.1;
- `libxkbcommon`, from version 1.3.1 to 1.4.1;
- `libxml`, from version 2.9.12 to 2.10.0;
- `links`, from version 2.25 to 2.27;
- `nettle`, from version 3.7.3 to 3.8.1.

The dockerfile got updated to install `python3-xcbgen` which is needed for `libxcb` to generate some files during compile time.
`fontconfig` got a missing patch fixed and added, this should unbreak ci.
Lastly, `xbps` got a revbump because the recent libarchive update seems to have bumped sonames.

This is not everything but I'm breaking it up into smaller parts due to vacation.

### DIFF
--- a/bootstrap.d/dev-libs.yml
+++ b/bootstrap.d/dev-libs.yml
@@ -1076,8 +1076,8 @@ packages:
     source:
       subdir: 'ports'
       git: 'https://git.lysator.liu.se/nettle/nettle.git'
-      tag: 'nettle_3.7.3_release_20210606'
-      version: '3.7.3'
+      tag: 'nettle_3.8.1_release_20220727'
+      version: '3.8.1'
       tools_required:
         - host-autoconf-v2.69
         - host-automake-v1.15
@@ -1094,7 +1094,6 @@ packages:
     pkgs_required:
       - mlibc
       - gmp
-    revision: 3
     configure:
       - args:
         - '@THIS_SOURCE_DIR@/configure'
--- a/bootstrap.d/dev-libs.yml
+++ b/bootstrap.d/dev-libs.yml
@@ -888,11 +888,11 @@ packages:
       categories: ['dev-libs']
     source:
       subdir: 'ports'
-      url: 'https://ftp.gnu.org/gnu/libunistring/libunistring-0.9.10.tar.xz'
+      url: 'https://ftp.gnu.org/gnu/libunistring/libunistring-1.0.tar.xz'
       format: 'tar.xz'
-      checksum: blake2b:25d162d9d510cc35ad4209acceb9b06bcc0553b8ce56e94f8df12c4df64d91abfc4a9e15b50b5c8d5b9672939305a394a7e83f1892258defb7ae5ac2ccf79dfb
-      extract_path: 'libunistring-0.9.10'
-      version: '0.9.10'
+      checksum: blake2b:8208fe33d4ac2f015b0efb56b0c7dd87afc4bb1c6ca4eb3ded86d7e2101d7b7f68bfd8991af4b6dd408282ec73f134ee0b884e761ff6d52e8a1e398326aec420
+      extract_path: 'libunistring-1.0'
+      version: '1.0'
       tools_required:
         - host-autoconf-v2.69
         - host-automake-v1.15
@@ -904,7 +904,6 @@ packages:
       - system-gcc
     pkgs_required:
       - mlibc
-    revision: 4
     configure:
       - args:
         - '@THIS_SOURCE_DIR@/configure'
@@ -912,7 +911,7 @@ packages:
         - '--prefix=/usr'
         - '--with-sysroot=@SYSROOT_DIR@' # Set libtool's lt_sysroot.
         - '--disable-static'
-        - '--docdir=/usr/share/doc/libunistring-0.9.10'
+        - '--docdir=/usr/share/doc/libunistring-1.0'
     build:
       - args: ['make', '-j@PARALLELISM@']
       - args: ['make', 'install']
--- a/bootstrap.d/dev-libs.yml
+++ b/bootstrap.d/dev-libs.yml
@@ -848,8 +848,8 @@ packages:
     source:
       subdir: ports
       git: 'https://gitlab.com/gnutls/libtasn1.git'
-      tag: 'v4.17.0'
-      version: '4.17.0'
+      tag: 'v4.18.0'
+      version: '4.18.0'
       tools_required:
         - host-autoconf-v2.69
         - host-automake-v1.15
@@ -864,7 +864,6 @@ packages:
       - system-gcc
     pkgs_required:
       - mlibc
-    revision: 3
     configure:
       - args:
         - '@THIS_SOURCE_DIR@/configure'
--- a/bootstrap.d/dev-libs.yml
+++ b/bootstrap.d/dev-libs.yml
@@ -815,8 +815,8 @@ packages:
     source:
       subdir: ports
       git: 'https://gitlab.com/cjwatson/libpipeline.git'
-      tag: '1.5.4'
-      version: '1.5.4'
+      tag: '1.5.6'
+      version: '1.5.6'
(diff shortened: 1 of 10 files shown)


bootstrap.d/dev-libs.yml
```diff
@@ -973,19 +968,16 @@ packages:
...
-      - python
       - libiconv
-    revision: 3
     configure:
       - args:
         - '@THIS_SOURCE_DIR@/configure'
         - '--host=@OPTION:arch-triple@'
         - '--prefix=/usr'
-        - '--with-python=@SYSROOT_DIR@/usr/bin/python3.8'
+        - '--without-python'
```
avdgrinten, line 980: Why is this done?

bootstrap.d/x11-libs.yml
```diff
@@ -608,7 +605,7 @@ packages:
         - '--without-doxygen'
         - '--with-sysroot=@SYSROOT_DIR@' # Set libtool's lt_sysroot.
         environ:
-          PYTHON: '@BUILD_ROOT@/tools/host-python/bin/python3.8'
+          PYTHON: '/usr/bin/python3'
```
avdgrinten, line 608: Why is this done? If we already have a host-python, why don't we use it?

docker/Dockerfile
```diff
@@ -83,6 +83,8 @@ RUN apt-get update \
 		python3-wheel \
 		# Used by managarm
 		python3-protobuf \
+		# Used by libxcb
+		python3-xcbgen \
```
avdgrinten, line 87: By how much does the rootfs size increase if we pull this package?

bootstrap.d/dev-libs.yml
```diff
@@ -627,8 +625,9 @@ packages:
         - host-gettext
       regenerate:
         - args: ['./gitsub.sh', 'pull']
-        # Gnulib broke on commit e3174b6d1fdbe6ea2297bf8c8333f65f9d9d9588, so check out the one before that.
-        - args: ['git', 'checkout', '766ec17a90f67e8cda78394e58a7fffb00f5a4b7']
+        # Checkout current gnulib master, 044bf893acee0a55b22b4be0ede0e3ce010c480a, 29-08-2022.
+        # This avoids random breakage in the future
+        - args: ['git', 'checkout', '044bf893acee0a55b22b4be0ede0e3ce010c480a']
```
avdgrinten, line 630: This also means that we won't get updated automatically though.

avdgrinten: Is it feasible to add gnulib as a source, so that we can use a local copy of gnulib that we can also update regularly?

bootstrap.d/x11-libs.yml
ArsenArsen: host-python seems a bit silly, but so is hardcoding this path if it's overriding $PATH

bootstrap.d/dev-libs.yml
ArsenArsen: that would be preferable

bootstrap.d/dev-libs.yml
Dennisbonke: Needs python2, we don't have that.

bootstrap.d/x11-libs.yml
Dennisbonke: Host-python doesn't have xcbgen.

docker/Dockerfile
Dennisbonke: Uhhhh, unsure, but I'm pretty sure it's small. I'll check when I have time and remember how to do that.

bootstrap.d/dev-libs.yml
Dennisbonke: While preferred, unlikely, as different packages want different versions.

bootstrap.d/x11-libs.yml
ArsenArsen: `PYTHON: python3` should also do that, no? also, isn't the default py3 now on debian?

bootstrap.d/dev-libs.yml
ArsenArsen: python is backwards compatible

ArsenArsen: oh, my bad, that's for the other comment threat; are you sure newest version doesn't work for all?

no92 (approved):

bootstrap.d/x11-libs.yml
Dennisbonke: Seems to work fine, I've changed it.

no92 (approved):
LGTM

docker/Dockerfile
Dennisbonke: Less than 10 megabytes, I don't have anything more accurate than that. Note that this is the output of `docker images`, so this is uncompressed.

bootstrap.d/dev-libs.yml
ArsenArsen: I don't remember whether I asked this before, but I think I did: does something wanting gnulib X break with gnulib Y (when Y>X)?

patches/fontconfig/0001-Add-Managarm-support.patch
```diff
@@ -0,0 +1,26 @@
+From 817dc2102610dd853a718e14eee63373ab31c8eb Mon Sep 17 00:00:00 2001
+From: Dennis Bonke <admin@dennisbonke.com>
+Date: Mon, 29 Aug 2022 01:04:27 +0200
+Subject: [PATCH] Add Managarm support
+
+Signed-off-by: Dennis Bonke <admin@dennisbonke.com>
+---
```
ArsenArsen, line 7: is there a matching revbump?

bootstrap.d/dev-libs.yml
Dennisbonke: It can and has done so in the past iirc.

ArsenArsen (commented):
looks almost there

also, I missed the comment on this, but: why is host-py removed in favor of debian-py in one of the packages

bootstrap.d/dev-libs.yml
ArsenArsen: in that case, seems more reasonable to stick to dist tarballs for anything that uses gnulib, but that's off topic for this PR

Dennisbonke:
> looks almost there
> 
> 
> 
> also, I missed the comment on this, but: why is host-py removed in favor of debian-py in one of the packages

That has to do with needing python3-xcbgen in that package.

bootstrap.d/dev-libs.yml
Dennisbonke: Yeah I was looking at that, but out of scope for this PR.

patches/fontconfig/0001-Add-Managarm-support.patch
Dennisbonke: Nope, and it isn't needed. Fontconfig got updated in #206 and never build successfully on the ci.

ArsenArsen: ah, fair enough

ArsenArsen (approved):

PR #210, opened 2022-10-03
Autotools rework

Dennisbonke:
This PR updates the autotool stack to the most recent versions (the old ones are still available for now, tho one should consider them deprecated). It also reworks our config.sub patch to match the one that was upstreamed by @ArsenArsen (pending approval from the upstream maintainer). While I was working on that, we handle autoreconf and friends differently, courtesy of @mintsuki for pointing that out. As I want the basework in asap, most packages will be converted to the new autotools stack in a future PR. As this PR touches the gcc patch, I'm holding off on upgrading the toolchain until this PR is in, but it's on the TODO list!

The following tools have been touched by this PR:
- `host-autoconf-v2.71`, tool added;
- `host-automake-v1.16`, tool added;
- `host-libtool`, updated from version 2.4.6 to 2.4.7 and reworked the patch;
- `cross-binutils`, reworked the way we regenerate it;
- All the `gcc` tools, made the patch work with the new config.sub, updated the regenerate steps, updated autotools stack.


bootstrap.yml
```diff
@@ -118,13 +146,39 @@ sources:
...
+        workdir: '@THIS_SOURCE_DIR@/libbacktrace'
+      - args: ['autoreconf', '-fvi', '-I../config']
+        workdir: '@THIS_SOURCE_DIR@/lto-plugin'
+      - args: ['libtoolize', '-cfvi']
+        workdir: '@THIS_SOURCE_DIR@/lto-plugin'
+      - args: ['autoreconf', '-fvi', '-I../config']
+        workdir: '@THIS_SOURCE_DIR@/libcc1'
+      - args: ['libtoolize', '-cfvi']
+        workdir: '@THIS_SOURCE_DIR@/libcc1'
+      - args: ['autoreconf', '-fvi', '-I../config']
```
ArsenArsen, line 178: Why repeat this instead of just overriding the config.sub autoconf holds?

Dennisbonke: According to mint, this is the better way of doing it. I tried not doing it in all directories but that will give libtool version mismatches.

ArsenArsen: what's `-I../config` for? my guess was to specify where config.{guess,sub} are, but we should just put them into whatever provides them instead

ArsenArsen: also autoreconf is broken on GCC, it has a script that can update files in the right order, but since we know what we patch, it's probably better to just invoke the tools directly in the specific case of GCC

ArsenArsen: yes, by coincidence, not design. we should definitely prefer to have no `-I` since that can only confuse whoever else is trying to use autoconf

bootstrap.yml
```diff
@@ -65,11 +65,39 @@ sources:
...
+      - args: ['libtoolize', '-cfvi']
+        workdir: '@THIS_SOURCE_DIR@/gas'
+      - args: ['autoreconf', '-fvi', '-I../config']
+        workdir: '@THIS_SOURCE_DIR@/binutils'
+      - args: ['libtoolize', '-cfvi']
+        workdir: '@THIS_SOURCE_DIR@/binutils'
+      - args: ['autoreconf', '-fvi', '-I../config']
+        workdir: '@THIS_SOURCE_DIR@/gprof'
+      - args: ['libtoolize', '-cfvi']
+        workdir: '@THIS_SOURCE_DIR@/gprof'
```
avdgrinten, line 100: Why is this done?

bootstrap.yml
```diff
@@ -118,13 +146,39 @@ sources:
...
+      - args: ['libtoolize', '-cfvi']
+        workdir: '@THIS_SOURCE_DIR@/lto-plugin'
+      - args: ['autoreconf', '-fvi', '-I../config']
+        workdir: '@THIS_SOURCE_DIR@/libcc1'
+      - args: ['libtoolize', '-cfvi']
+        workdir: '@THIS_SOURCE_DIR@/libcc1'
+      - args: ['autoreconf', '-fvi', '-I../config']
+        workdir: '@THIS_SOURCE_DIR@/libcpp'
+      - args: ['libtoolize', '-cfvi']
+        workdir: '@THIS_SOURCE_DIR@/libcpp'
```
avdgrinten, line 181: Why is this done?

bootstrap.yml
```diff
@@ -657,8 +761,8 @@ packages:
         # -g blows up the binary size.
         - 'CFLAGS=-O2'
     build:
-      #- args: ['make', '-j@PARALLELISM@', 'all-binutils', 'all-gas', 'all-ld']
-      - args: ['make', 'all-binutils', 'all-gas', 'all-ld']
+      - args: ['make', '-j@PARALLELISM@', 'all-binutils', 'all-gas', 'all-ld']
+      # - args: ['make', 'all-binutils', 'all-gas', 'all-ld']
```
avdgrinten, line 765: We can drop the comment.

patches/gcc/0001-Add-Managarm-target.patch
```diff
@@ -14,69 +13,25 @@ TODO: Use libtoolize instead of modifying in-tree libtool?
...
-diff --git a/config.sub b/config.sub
-index 63c1f1c8b..033763b58 100755
---- a/config.sub
-+++ b/config.sub
-@@ -132,7 +132,8 @@ case $1 in
- 		maybe_os=$field2-$field3
- 		case $maybe_os in
- 			nto-qnx* | linux-* | uclinux-uclibc* \
--			| uclinux-gnu* | kfreebsd*-gnu* | knetbsd*-gnu* | netbsd*-gnu* \
-+			| uclinux-gnu* | managarm-kernel* | managarm-system* \
```
avdgrinten, line 46: Does the upstream Managarm target use `-system` or `-mlibc`? Is the plan to switch to `-mlibc`?

patches/gcc/0001-Add-Managarm-target.patch
```diff
@@ -105,7 +60,7 @@ index 357b0bed0..570049be8 100644
 +  use_gcc_stdint=wrap
 +
 +  case ${target} in
-+  *-managarm-system*)
++  *-managarm-mlibc*)
```
avdgrinten, line 63: I'm confused. We use `-system` in config.sub but here we replace the existing `-system` patch by `-mlibc`?

ArsenArsen: config.sub upstream has -mlibc, so we should be going for that

Dennisbonke: Agreed

patches/gcc/0001-Add-Managarm-target.patch
Dennisbonke: Yes, that is what landed in config.sub upstream.

bootstrap.yml
Dennisbonke: Will do.

bootstrap.yml
Dennisbonke: According to mint, this is the better way of doing it. I tried not doing it in all directories but that will give libtool version mismatches. See also [this](https://github.com/managarm/bootstrap-managarm/pull/210#discussion_r986438187) comment thread.

ArsenArsen: this is another reason to invoke the right version of the right tool, and that alone, for regenerating gcc files (remember, we aren't dealing with a normal autotools build system, but a broken one)

Dennisbonke: Yeah not gonna carry libtool 2.2.7, gcc can suck it.

ArsenArsen: didn't say that you should, neither autoconf and automake install libtool, this is only a problem if you don't use those (on top of the other problems that using wrong versions will cause)

ArsenArsen (changes requested):
also (I didn't see this before), we should drop any hunks that touch config.sub in favor of just replacing config.sub outright from a single source (it's far cleaner that way)

Dennisbonke:
> also (I didn't see this before), we should drop any hunks that touch config.sub in favor of just replacing config.sub outright from a single source (it's far cleaner that way)

I will be working on that after this is merged, checking all ports and working through them if needed.

patches/gcc/0001-Add-Managarm-target.patch
avdgrinten: But if we do that, we should also use `-mlibc` in config.sub. I'm confused by the mismatch.

Dennisbonke:
I think we reached a consensus here. Can this be approved and merged? It's blocking me upgrading the toolchain.

ArsenArsen:
I recall no such consensus, use your best judgement and feel free to merge

Dennisbonke:
Let me ask that differently, apart from your concerns about gcc/binutils and autotools versions, can you (and anyone else for that matter) approve of the rest of this PR?

PR #211, opened 2022-10-08
Add several ports I had laying around

Dennisbonke:
All of these ports are split off from #178, and as far as I can tell should just be good to go.

- `snappy`;
- `libx264`;
- `opus`;
- `libvpx`;
- `ffmpeg`;
- `lcms`;
- `re2`.


bootstrap.d/media-video.yml
```diff
@@ -0,0 +1,107 @@
...
+        - '--disable-v4l2-m2m'
+        - '--disable-vaapi'
+        - '--disable-vdpau'
+        - '--disable-videotoolbox'
+        # Start cross compilation config
+        - '--enable-cross-compile'
+        - '--arch=@OPTION:arch@'
+        - '--cross-prefix=@OPTION:arch-triple@'
+        - '--host-cc=gcc'
+        - '--target-os=linux' # Fun times ahead
```
avdgrinten, line 85: Why don't we add a target for Managarm?

patches/ffmpeg/0001-Add-initial-Managarm-support.patch
```diff
@@ -0,0 +1,26 @@
...
+diff --git a/libavutil/cpu.c b/libavutil/cpu.c
+index 6bd0f07..4d692e4 100644
+--- a/libavutil/cpu.c
++++ b/libavutil/cpu.c
+@@ -280,7 +280,7 @@ int av_cpu_count(void)
+ #if HAVE_WINRT
+     SYSTEM_INFO sysinfo;
+ #endif
+-#if HAVE_SCHED_GETAFFINITY && defined(CPU_COUNT)
++#if HAVE_SCHED_GETAFFINITY && defined(CPU_COUNT) && !defined(__managarm__)
```
avdgrinten, line 20: Why is this done?

bootstrap.d/media-video.yml
Dennisbonke: Probably me being lazy and it seems to work out fine, so no need to add extra patchwork that needs to be upstreamed eventually.

patches/ffmpeg/0001-Add-initial-Managarm-support.patch
Dennisbonke: We don't implement this yet (or at least we didn't at the time of me porting ffmpeg).

no92: mlibc has it stubbed, but managarm currently does not implement this. This would have to be implemented as a new hel call IIRC.

no92 (approved):

patches/ffmpeg/0001-Add-initial-Managarm-support.patch
avdgrinten: We should add the required infrastructure to mlibc (if it doesn't exist already) and make the Managarm sysdeps return ENOSYS (and log a message).

PR #214, opened 2022-10-19
Add a cmatrix port

Dennisbonke:
This PR adds a `cmatrix` port, enables `sl` for AArch64 and fixes some metadata ordering.


Dennisbonke:
Both `sl` and `cmatrix` build fine on AArch64.

PR #215, opened 2022-10-19
nyancat: Add missing metadata and build on AArch64

Dennisbonke:
Can't leave AArch64 without nyancat can we? Untested but I see no issues with this build.


avdgrinten:
This should be easy to test before its merged.

no92 (approved):

PR #224, opened 2023-02-07
scripts: add a task for testing mlibc sysdeps like on CI

no92:


scripts/compile-mlibc-sysdeps.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+SYSDEPS=${3:-"dripos lemon aero ironclad vinix lyre"}
+cd pkg-builds/
+for sysdep in $(echo $SYSDEPS); do
```
ArsenArsen, line 5: redundant subshell

scripts/compile-mlibc-sysdeps.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+SYSDEPS=${3:-"dripos lemon aero ironclad vinix lyre"}
```
ArsenArsen, line 3: it's prob better to just pop the non-sysdeplist arguments first and use $@ as the list of subsysdeps

scripts/compile-mlibc-sysdeps.sh
```diff
@@ -0,0 +1,16 @@
...
+cd pkg-builds/
+for sysdep in $(echo $SYSDEPS); do
+	if [ ! -d "mlibc-sysdep-$sysdep" ]; then
+		meson setup \
+			"-Dc_args=['-fno-stack-protector', '-U__linux__']" \
+			"-Dcpp_args=['-fno-stack-protector', '-U__linux__']" \
+			-Dbuild_tests=true \
+			-Db_sanitize=undefined \
+			"-Dlinux_kernel_headers=$2" \
+			--cross-file $1/ci/$sysdep.cross-file mlibc-sysdep-$sysdep $1
```
ArsenArsen, line 13: quoting is missing, also prefer ${} for cases like $foo.bar, since that's a lot more readable in such contexts

ArsenArsen (changes requested):
I'm not sure this should be in bootstrap-managarm, but I do like the idea.

scripts/compile-mlibc-sysdeps.sh
no92: This doesn't really work, as `$@` is an array IIRC. Maybe this is better?
```bash
SYSDEPS=${*:-"dripos lemon aero ironclad vinix lyre"}
```
The shifting doesn't help there, too, but is nice for the sake of clarifying what `$1` and `$2` mean.

no92:
Addressed the issues raised. Also updated the path for the kernel headers that are passed to the script.

scripts/compile-mlibc-sysdeps.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
ArsenArsen, line 1: this uses no bash features, it can be /bin/sh

also use shellcheck

scripts/compile-mlibc-sysdeps.sh
```diff
@@ -0,0 +1,18 @@
...
+MLIBC_PATH=$1
+KERNEL_HEADER_PATH=$2
```
ArsenArsen, line 3: missing quoting, use lowercase for non-export variables

scripts/compile-mlibc-sysdeps.sh
```diff
@@ -0,0 +1,18 @@
...
+shift 2
+SYSDEPS=${*:-"dripos lemon aero ironclad vinix lyre"}
```
ArsenArsen, line 5: redundant variable, just use "$@" in the for loop, or just `for x; do`, since that iterates over $@ anyway

ArsenArsen (changes requested):
CI should probably not duplicate this logic either

scripts/compile-mlibc-sysdeps.sh
ArsenArsen: SYSDEPS is redundant

ArsenArsen (approved):

Dennisbonke (approved):
LGTM

PR #226, opened 2023-02-08
mesa: disable futexes mistakenly enabled by having linux/futex.h

no92:


Dennisbonke (changes requested):
LGTM but please bump the revision of `mesa` too.

ArsenArsen:
I think we should start documenting what patches fix too.  Could you add a bit of commentary below the `---` line?

no92:
Addressed the comments raised above.

Dennisbonke (approved):
LGTM

PR #232, opened 2023-04-28
dev-libs: Add freestanding-headers port

mintsuki:
This PR complements mlibc#859


no92 (approved):
I'm not sure on whether the `freestanding-headers` package should go into `dev-libs` or just the root `bootstrap.yml` alongside with mlibc, but that's a very minor concern to me TBH.

LGTM.

PR #234, opened 2023-05-01
update the GNU toolchain, remove cxxshim

ArsenArsen:
requires mlibc and managarm changes, merge after https://github.com/managarm/managarm/pull/531 and https://github.com/managarm/mlibc/pull/863

builds most of the distro locally on xbbs, there's some -Werror (mis)use that breaks pango and a few other packages

multilib was dropped from userspace gcc.

also features a QoL change to `vm-utils.py gdb`


no92 (approved):
LGTM

Dennisbonke (commented):
I'm pretty sure that the complete removal of the gcc and binutils patches is broken. Unless we upstreamed all of it, which I'm not aware of?

ArsenArsen:
it is not and we did not, the sources contain them

Dennisbonke:
Oh I see yeah. Looks good to me then.

Dennisbonke (approved):
LGTM

PR #239, opened 2023-07-18
gtklife: update checksum, use https

no92:
Apparently, the checksum for the archive changed despite not receiving a version bump. Checking with gpg against the authors public key is fine, though, so we need to update our blake2b hash.


Dennisbonke (approved):
LGTM

PR #241, opened 2023-07-19
libdrm: rebase patches and enable linux codepaths for managarm

no92:
This allows for successfully running `drmGetDevice` et al. when the necessary sysfs changes land in managarm. Without them, the function will still fail with an error, same as before.


Dennisbonke (approved):
LGTM

PR #259, opened 2023-08-20
bootstrap, sys-firmware: Add various linux-firmware packages

ElectrodeYT:


no92 (approved):
LGTM

qookei:
Why do we need linux-firmware on managarm?

ikbenlike:
One nitpick: amd-ucode is the only package not prefixed with `linux-firmware`. Is there a particular reason for this?

ElectrodeYT:
> Why do we need linux-firmware on managarm?

I plan on attempting to bring up a Radeon card, which would require firmware files from the `amdgpu` (and maybe the `radeon`) folder.
(For context, the nvidia card I was writing a driver for broke, so that plan has died)

In addition, there really is no point in not adding packages for other drivers such as i915. @no92 is also working on a intel GPU driver, which _may_ require firmware files from the `i915` folder.

The amd-ucode package was made simply because I could, and loading microcode at some point would probably be a good idea :meme:

ElectrodeYT:
> One nitpick: amd-ucode is the only package not prefixed with `linux-firmware`. Is there a particular reason for this?

Convention; most distributions do this, probably because the intel equivalent (which is not made with linux-firmware) is called `intel-ucode`

Dennisbonke:
Legality question, can we redistribute this as a binary package? Or would that pose issues? @ArsenArsen @avdgrinten

ElectrodeYT:
> Legality question, can we redistribute this as a binary package? Or would that pose issues? @ArsenArsen @avdgrinten

As far as I know, it is presumed that all (or at least most) files in linux-firmware are at least redistributable.
The files I picked out with the packages are all redistributable.

ArsenArsen:
some of them, see https://gitweb.gentoo.org/repo/gentoo.git/tree/sys-kernel/linux-firmware/linux-firmware-20230804.ebuild (specifically, stuff affected by the bindist and redistributable flags).

ElectrodeYT:
Bump here again, an intel-lil GPU driver would benefit from having the GuC and HuC firmware blobs, which we would need to supply. Loading CPU microcode would also be interesting to do at some point.

As mentioned before, all packages created here only contain firmware files which are redistributable.

PR #260, opened 2023-10-02
Add support for generating docs

no92:


bootstrap.yml
```diff
@@ -444,6 +444,21 @@ tools:
       - args: ['ninja', 'install']
         quiet: true
 
+  - name: host-xxd
```
Dennisbonke, line 447: Missing labels and the architecture field.

bootstrap.yml
```diff
@@ -686,6 +701,29 @@ tools:
     install:
       - args: ['ninja', 'install']
 
+  - name: host-hdoc
```
Dennisbonke, line 704: Missing labels and the architecture field.

Dennisbonke (changes requested):
Two items that will break xbbs otherwise. Outside that, LGTM

no92:
Addressed the issues raised.

Dennisbonke (approved):
LGTM

PR #261, opened 2023-10-02
vm-util: add option for remote target ip with gdb

no92:


ArsenArsen:
why not just replace the whole remote string instead?  that way, the port and protocol are also controllable

no92:
The script hardcodes these parameters for qemu anyways, so the use-case for that would be significantly more limited.

Dennisbonke (approved):
LGTM

PR #263, opened 2023-10-04
wlgears: add port

no92:


bootstrap.d/gui-apps.yml
```diff
@@ -45,3 +45,45 @@ packages:
...
+      summary: A port of glxgears to wayland
+      description: "This utility is a port of the glxgears demo of mesa-demos onto wayland with a meson build system."
+      spdx: MIT
+      website: 'https://github.com/no92/wlgears'
+      maintainer: "no92 <no92.mail@gmail.com>"
+      categories: ['gui-apps']
+    source:
+      subdir: ports
+      git: 'https://github.com/no92/wlgears.git'
+      branch: master
```
Dennisbonke, line 61: Ideally we pin a commit, which can be overlooked because you control the repo, but please do pin a commit.
You're also lacking a version tag, which is required.

Dennisbonke (changes requested):

bootstrap.d/gui-apps.yml
no92: Fixed.

Dennisbonke (approved):
LGTM

PR #270, opened 2023-10-12
sysstat: add port

ElectrodeYT:
Mainly useful for `mpstat`, which (combined with my WIP scheduler PR) can read out the current CPU and boot-time average CPU usage from `/proc/stat`.


Dennisbonke (approved):
LGTM

PR #273, opened 2023-10-15
General updates, part 1

ElectrodeYT:


bootstrap.d/media-fonts.yml
```diff
@@ -2,8 +2,8 @@ sources:
   - name: xorg-font-util
     subdir: 'ports'
     git: 'https://gitlab.freedesktop.org/xorg/font/util.git'
-    tag: 'font-util-1.3.2'
-    version: '1.3.2'
+    tag: 'font-util-1.4.0'
+    version: '1.4.0'
```
Dennisbonke, line 6: You should also reset the revision on `host-xorg-font-util`/

bootstrap.d/dev-libs.yml
```diff
@@ -1498,8 +1498,8 @@ packages:
     source:
       subdir: ports
       git: 'https://github.com/PhilipHazel/pcre2.git'
-      tag: 'pcre2-10.39'
```
Dennisbonke, line 1504: Update the docdir too

Dennisbonke (changes requested):
A few items have caught my eye. Furthermore, all the commit messages are wrong. It should be `<package name>: Update from x to y` or `<package name>: Update to version x`. Initially we used the first version but I'm starting to lean towards the second one, so use that one. Please reword all commit messages to follow that.

Dennisbonke (approved):
LGTM now

PR #275, opened 2023-10-18
openttd-openmsx: fix checksum

ElectrodeYT:
Sorry that this happened! Looks like I somehow managed to not change this checksum when I was creating this recipe


Dennisbonke (approved):
LGTM

PR #286, opened 2023-11-20
mesa-git: add port

ElectrodeYT:
This package is effectively a rolling-release version of mesa, tracking the upstream main branch.

This port does not require any patches, due to the recent up streaming of our patches into mainline mesa.

### DIFF
--- a/bootstrap.d/media-libs.yml
+++ b/bootstrap.d/media-libs.yml
@@ -871,6 +871,93 @@ packages:
           DESTDIR: '@THIS_COLLECT_DIR@'
         quiet: true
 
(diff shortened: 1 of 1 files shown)


Dennisbonke (changes requested):
NAK, I'd rather hold for the next proper release

ElectrodeYT:
I was not aware that this would make xbps build every commit, I was mostly thinking about this as something that one can build locally.

Would it be fine if I would add `stability_level: 'broken'` to make xbps not build it?

Dennisbonke:
No longer needed

PR #291, opened 2023-12-18
Limine 6 fixes

qookei:
Update the disk layout to accomodate Limine 6 dropping support for Ext2/3/4.

This will require everyone to create new disk images since the ESP needs to be
larger now. `update-image.py` will issue an

Depends on https://github.com/qookei/image_create/pull/1 for creating images
with the new partition table layout.

### DIFF
--- a/scripts/update-image.py
+++ b/scripts/update-image.py
@@ -10,13 +10,14 @@ import shlex
 import shutil
 import subprocess
 import sys
+import tempfile
 import time
 import traceback
 import uuid
 from dataclasses import dataclass, field
 from enum import Enum
 from os import path
-from typing import Dict
+from typing import Dict, List
 
 elevation_method = None
 verbose = False
@@ -38,10 +39,11 @@ class MountInfo:
 
 
 class Partition:
-    def __init__(self, idx, type, uuid):
+    def __init__(self, idx, type, uuid, size):
         self.idx = idx
         self.type = type
         self.uuid = uuid
+        self.size = size
 
 
 # shlex.join is only available since Python 3.8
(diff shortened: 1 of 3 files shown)


ElectrodeYT:
Does this require init changes in posix? As far as I can tell, stage1 does not at present mount `/boot`.

qookei:
We don't care about `/boot` or `/boot/efi` in init, so no.

Dennisbonke (approved):
LGTM

ElectrodeYT (approved):
I'll save my opinions on this limine change here, but if we must update to limine 6, this is OK.

scripts/update-image.py
```diff
@@ -801,14 +801,20 @@ def setup_esp(start, size):
 
             # manipulate it using mtools
             dosimg = "{}@@{}".format(self.image, start * 512)
-            # create \EFI
+            # create \EFI\BOOT
             logged_check_call(["mmd", "-i", dosimg, "EFI"])
+            logged_check_call(["mmd", "-i", dosimg, "EFI/BOOT"])
+            # HACK(qookie): We add CREATE_DIR tasks for these two...
+            # create \managarm
```
ArsenArsen, line 808: hmm, why?

scripts/update-image.py
```diff
@@ -895,7 +901,7 @@ def run(self):
     "-m",
     "--mount-using",
     dest="mount_using",
-    choices=["loopback", "docker", "guestfs", "block", "remake"],
+    choices=["loopback", "docker", "guestfs", "block"],
```
ArsenArsen, line 904: where did remake go?

qookei: remake is an action, not a way to mount the image (trying to do `-m remake` fails in the mount action) :^)

scripts/update-image.py
qookei: We create `\managarm` and `\grub` because files get copied there (the kernel & initrd, and `grub.cfg` respectively), which fails if they don't exist.

PR #299, opened 2024-02-05
system-gcc: Disable decimal float support

Qwinci:
Disable gcc decimal float support because the lack of emulation routines for it in our AArch64 libgcc build causes issues in mpfr. This shouldn't affect most other packages in any way because decimal floats aren't used on most of them.

### DIFF
--- a/bootstrap.yml
+++ b/bootstrap.yml
@@ -584,6 +584,7 @@ tools:
         - '--enable-languages=c,c++,lto'
         - '--enable-initfini-array'
         - '--enable-libstdcxx-filesystem-ts'
+        - '--disable-decimal-float'
         # -g blows up GCC's binary size.
         - 'CFLAGS=-O2'
         - 'CXXFLAGS=-O2'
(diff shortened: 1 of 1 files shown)


bootstrap.yml
```diff
@@ -584,6 +584,7 @@ tools:
         - '--enable-languages=c,c++,lto'
         - '--enable-initfini-array'
         - '--enable-libstdcxx-filesystem-ts'
+        - '--disable-decimal-float'
```
Dennisbonke, line 587: We'd need to bump the revision on this tool.

ArsenArsen:
which issues?

ArsenArsen:
a proper fix would be rather simple - just adding `t-dfprules` to tmake_file in the `aarch64*-*-managarm*)` section of `config.host`.  if someone prepares a patch for our gcc fork, that'd be nice - we can roll a release about it after.  does this issue happen on other arches?  I can see that it's there in the x86_64 case at least

PR #307, opened 2024-02-15
Initial wine port

Qwinci:
Add initial wine port, initial recipe work for host-wine-tools/wine originally done by @Dennisbonke.


bootstrap.d/app-emulation.yml
```diff
@@ -69,3 +139,89 @@ packages:
...
+      - libxrender
+      - libxrandr
+      - libxshmfence
+      - libxxf86vm
+      - libjpeg-turbo
+      - libpng
+      - libxml
+      - libxslt
+      - zlib
+      # - glib
```
no92, line 177: What's the background for these commented-out dependencies?

no92 (approved):
LGTM, minor questions remain.

bootstrap.d/app-emulation.yml
Qwinci: Doesn't look like they are directly needed by the latest wine so I removed them and updated the dependencies in general.

PR #327, opened 2024-04-08
Drop protobuf

no92:
As managarm/managarm#652 is merged now, we can drop these. The patches are horribly out of date and hard to rebase, so nothing is really lost here.


qookei (approved):

Dennisbonke (approved):
LGTM

PR #330, opened 2024-04-12
host-managarm-tools: bump revision

no92:
This is necessary to update the xbbs-built version, as it continues to rely on protobuf, which we recently removed.


Dennisbonke (approved):
LGTM

PR #342, opened 2024-05-24
dhcpcd: Add port

no92:
Depends on managarm/managarm#662 and managarm/mlibc#1062.


Dennisbonke:
This has been pushed into master now. Thanks for the port!

PR #347, opened 2024-06-02
libdrm-headers: Add package

no92:
This package is derived from the libdrm source, and only serves to extract the drm headers into a package. mlibc depends on having libdrm headers present for building managarm sysdeps, and libdrm does not support extracting the headers. This commit adds the libdrm-headers package, which copies out the headers and supplies a `meson.build` to allow for other meson-based projects (like mlibc) to use the headers as a subproject. This avoids sandbox violations and hardcoding paths between packages.


Dennisbonke (approved):
LGTM

PR #348, opened 2024-06-02
scripts/vm-util: support USB passthrough with pcap capture and aliases

no92:
Device aliases for some supported USB devices are provided for convenience. The `--usb-passthrough-pcap` option records a pcap file, which is named like the input to the option.


Dennisbonke (approved):
LGTM

PR #351, opened 2024-06-12
managarm-system: use `mold` instead of `ld`

no92:


no92:
Depends on managarm/managarm#683.

ArsenArsen:
(not against the change.. in fact, i have already messed with mold before, just curious)

what's the reason for this change?

ElectrodeYT:
> (not against the change.. in fact, i have already messed with mold before, just curious)
> 
> what's the reason for this change?

It's quite a bit faster, and reduces iteration time by a decent amount, especially on slower machines, if a change is made that requires relinking several servers or other binaries.

bootstrap.d/sys-devel.yml
```diff
@@ -61,6 +61,26 @@ tools:
     install:
       - args: ['make', 'install']
 
+  - name: host-mold
```
Dennisbonke, line 64: I miss the label tag (if applicable) and the architecture flag. Can you add them?

Dennisbonke (changes requested):
One issue

no92:
Fixed the issue and rebased onto master.

Dennisbonke (approved):
LGTM

PR #352, opened 2024-06-12
`scripts/vm-util.py` additions

no92:


Dennisbonke (approved):
LGTM

scripts/vm-util.py
```diff
@@ -277,6 +278,10 @@ def do_qemu(args):
         for device in args.usb_passthrough_pcap:
             qemu_args += qemu_process_usb_passthrough(device, True)
 
+    qemu_sudo = []
+    if args.sudo:
+        qemu_sudo = ["sudo"]
```
ArsenArsen, line 283: launching graphical programs via sudo usually breaks, especially on wayland, and, more importantly, i'm not sure how good of an idea that is

KSPAtlas: > launching graphical programs via sudo usually breaks, especially on wayland, and, more importantly, i'm not sure how good of an idea that is

Aren't there use cases for running QEMU with sudo, such as some networking configurations?

ArsenArsen: I suspect it's never necessary - I'd need more details

no92:
Dropped the commit adding `--sudo`. Instead, we check that the user has set up the passthrough device for rootless passthrough, and error out if not.

scripts/vm-util.py
ArsenArsen: I doubt it was unavoidable.  we'd probably benefit more from supporting libvirt than running graphical programs with root, have you tried that?

PR #364, opened 2024-07-06
ports/util: add combined tzdata and tzcode package

apache-hb:
iana provides an amalgamated tzcode and tzdata package, which is useful since xbstrap cant create a single package from multiple sources. This was added to aid in cross compiling postgresql - which requires zic from tzcode as well as tzdata.

### DIFF
--- a/bootstrap.d/dev-util.yml
+++ b/bootstrap.d/dev-util.yml
@@ -53,6 +53,23 @@ tools:
     install:
       - args: ['make', 'install']
 
+  - name: host-tzdb
+    architecture: '@OPTION:arch@'
+    source:
+      subdir: 'ports'
+      url: 'https://data.iana.org/time-zones/releases/tzdb-2024a.tar.lz'
+      format: 'tar.xz'
+      checksum: blake2b:91d6c55c4f860ced760df84a1751174f1eb8ea83939b44d3a206dccbc9bac1713fbcde70391deaa4ed01d95e6de08bc0b751d296bebdf5d9ca0b4534942694cf
+      extract_path: 'tzdb-2024a'
+      version: '2024a'
+    configure:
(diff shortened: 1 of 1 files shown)


bootstrap.d/dev-util.yml
```diff
@@ -53,6 +53,25 @@ tools:
     install:
       - args: ['make', 'install']
 
+  - name: host-tzdb-2024a
```
Dennisbonke, line 56: I propose we name it host-tzdb. The version 2024a seems redundant here and means we need to update all users if we grab a new version.

bootstrap.d/dev-util.yml
```diff
@@ -53,6 +53,25 @@ tools:
...
+    architecture: '@OPTION:arch@'
+    source:
+      subdir: 'ports'
+      url: 'https://data.iana.org/time-zones/releases/tzdb-2024a.tar.lz'
+      format: 'tar.xz'
+      checksum: blake2b:91d6c55c4f860ced760df84a1751174f1eb8ea83939b44d3a206dccbc9bac1713fbcde70391deaa4ed01d95e6de08bc0b751d296bebdf5d9ca0b4534942694cf
+      extract_path: 'tzdb-2024a'
+      version: '2024a'
+    tools_required:
+      - system-gcc
```
Dennisbonke, line 66: Considering it's a tool, do we really need the compiler that targets managarm here?

Dennisbonke (changes requested):
2 items

bootstrap.d/dev-util.yml
apache-hb: fair

bootstrap.d/dev-util.yml
apache-hb: I wasnt sure, but removing it didnt break it. fixed

qookei (approved):

Dennisbonke (approved):
LGTM

PR #372, opened 2024-08-23
host-qemu: enable `-march=native`

no92:
We can do this due to host-qemu being containerless and linking against a lot of system libraries, making it highly unlikely to be portable between systems anyways.


Dennisbonke (approved):
LGTM

PR #376, opened 2024-09-05
Add a port of shadow and use it in kmscon

Dennisbonke:
This PR adds a port of shadow, and changes kmscon to use `/usr/bin/login` over `/usr/bin/bash`.

Right now, there are two users one can login to, their credentials are as follows:
user:password
`root`:`toor`
`managarm`:`managarm`

Blocked on managarm/managarm#708


no92 (approved):
LGTM

PR #378, opened 2024-09-07
bootstrap: enable freestnd-c{,xx}-hdrs for host-mlibc

qookei:
For managarm/mlibc#1122.


qookei:
No longer needed, the logic in the mlibc PR was changed to enable the headers automatically if the compiler target triple is wrong.

PR #380, opened 2024-09-17
x11-libs: Bump xcb-util{,-keysyms,-render-util} to fix CI

qookei:
The git submodules referenced by these packages changed locations, switch to latest commits and bump revisions to fix CI.


Dennisbonke (approved):
LGTM

PR #393, opened 2024-10-31
bootstrap: Run {initialize-empty,remake}-image tasks in the container

qookei:
Also make host-limine not a containerless tool and build it using host-llvm-toolchain.

The make-image task is still containerless because it actually mounts the disk image (either via loopback or guestfs), but ideally it shouldn't be, because it invokes 'strip' from cross-binutils, which just happens to work outside the container.

Depends on https://github.com/qookei/image_create/pull/2.


no92 (approved):
LGTM

avdgrinten (approved):

PR #397, opened 2024-11-03
Provide thor via initrd, drop multiboot1

no92:
Counterpart to managarm/managarm#734.

### DIFF
--- a/scripts/vm-util.py
+++ b/scripts/vm-util.py
@@ -3,10 +3,12 @@
 import argparse
 import os
 import shlex
+import shutil
 import subprocess
 import string
 import struct
 import sys
+import tempfile
 
 main_parser = argparse.ArgumentParser()
 main_subparsers = main_parser.add_subparsers()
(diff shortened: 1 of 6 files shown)


scripts/vm-util.py
```diff
@@ -346,7 +330,6 @@ def do_qemu(args):
 qemu_parser.add_argument("--no-kvm", action="store_true")
 qemu_parser.add_argument("--virtual-cpu", action="store_true")
 qemu_parser.add_argument("--no-smp", action="store_true")
-qemu_parser.add_argument("--virtual-boot", action="store_true")
```
avdgrinten, line 349: Can't we do `--virtual-boot` with MB2?

no92: qemu's `-kernel` only supports multiboot1.

avdgrinten: Uh, that's annoying. That might be a reason to keep MB1 support.

no92: The problem with multiboot1 support is that it will, by definition, never give us a RSDP when booting in UEFI-mode. It only works with BIOS, and maybe CSM (not sure about that).

ArsenArsen: FWIW it also supports the Linux boot protocol IIRC

it might also be relatively easy to add mb2 support

no92: I guess we could also just set up EFI vars to boot `eir-uefi`?

no92:
Updated the PR to rework `--virtual-boot` to work without eir-mb1. This PR is blocked on limine-bootloader/limine#428.

no92:
The virtual-boot configuration also relies on qookei/image_create#3.

PR #402, opened 2024-11-03
Fix {re,}make-image tasks producing broken BIOS images

qookei:
Requires a new xbstrap release


no92 (approved):
LGTM, good for merge when the next xbstrap release is tagged.

PR #403, opened 2024-11-08
scripts/vm-util: add `--device-spec` option

no92:
This allows for specifying how devices should be laid out in a YAML file. Currently, this implementation only supports PCI bridges and (multifunction) PCI devices, but can be extended to other devices if needed.

Example spec file:
```yaml
- name: nvidia-bridge
  type: pci-bridge
  devices:
    - name: nvidia-passthrough
      type: pci-multifunction-device
      functions:
        - host: "01:00.0"
        - host: "01:00.1"
        - host: "01:00.2"
        - host: "01:00.3"
- name: i226-V
  type: pci-device
  host: "04:00.0"
```

This passes through the multifunction device from the host at segment 0, bus 1, device 0, functions 0 through 3 to the guest, placing them behind a PCI bridge under one device. The PCI device is passed through and most likely placed on bus 0 by qemu.

### DIFF
--- a/scripts/vm-util.py
+++ b/scripts/vm-util.py
@@ -10,6 +10,7 @@ import string
 import struct
 import sys
 import tempfile
+import yaml
 
 main_parser = argparse.ArgumentParser()
 main_subparsers = main_parser.add_subparsers()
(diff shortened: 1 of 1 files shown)


avdgrinten:
We should also support matching devices based on PCI vendor/device ID.

avdgrinten:
Also, it may to convenient to be able to passthrough an entire device (i.e., all functions at the same time).

scripts/vm-util.py
```diff
@@ -82,6 +83,52 @@ def qemu_process_usb_passthrough(device, pcap):
 
     return ["-device", devstr]
 
+qemu_pci_bridges = 0
+
+def qemu_parse_device_spec(yml):
+    args = []
+
+    def parse_device(yml, **kwargs):
```
avdgrinten, line 91: Not sure if I like the way that this function exploits `**kwargs`.

PR #410, opened 2024-11-17
automake: Take config.sub from gnuconfig

avdgrinten:
This ensures that we use the newest `config.sub` and `config.guess`. Without this patch, `riscv64-managarm` is not recognized by automake.


no92 (approved):
LGTM even though we usually format source_required lists differently.

Dennisbonke (approved):
Agreed with Leo, but not a blocker
IMHO. LGTM

PR #418, opened 2024-11-24
dev-libs: upgrade libtool to 2.5.4 and remove patches as we're upstream now

mintsuki:


Dennisbonke (changes requested):
We should reset the revision of host-libtool in managarm-build.y4.yml, other than that thanks for the PR!

Dennisbonke (approved):
LGTM

PR #392, opened 2024-10-31
Various updates

Dennisbonke:
This PR updates the following packages:

- `less` to version 668;
- `libnl` to version 3.11.0;
- `findutils` to version 4.10.0;
- `gawk` to version 5.3.1;
- `gdbm` to version 1.24;
- `coreutils` to version 9.5.

It also adds missing metadata to the above packages and `sed`.

PR #421, opened 2024-12-17
A whole bunch of updates

Dennisbonke:
Potentially also fixes `libpng`, `icu` and `openssl` on CI, might break the QT stack but we'll see!

This PR updates the following packages:

- `less` to version 668;
- `libnl` to version 3.11.0;
- `findutils` to version 4.10.0;
- `gawk` to version 5.3.1;
- `gdbm` to version 1.24;
- `coreutils` to version 9.5.

It also adds missing metadata to the above packages and `sed`.

Furthermore, it updates the following packages:
- `libpng` to version 1.6.44;
- `icu` to version 76.1;
- `tzdata` to version 2024b;
- `harfbuzz` to version 10.1.0;
- `pcre2` to version 10.44;
- `iso-codes` to version 4.17.0;
- `gstreamer` to version 1.24.10;
- `gst-plugins-base` to version 1.24.10;
- `gst-plugins-good` to version 1.24.10;
- `gst-plugins-bad` to version 1.24.10;
- `libxkbcommon` to version 1.7.0;
- `xkeyboard-config` to version 2.43;
- `libwebp` to version 1.4.0;
- `mtdev` to version 1.1.7;
- The entire `Qt6` stack to version 6.8.1;
- `libxtrans` to version 1.5.2;
- `xorg-util-macros` to version 1.20.2;
- `xorg-proto` to version 2024.1;
- `xcb-proto` to version 1.17.0;
- `libxau` to version 1.0.12;
- `libxdmcp` to version 1.1.5;
- `libxcb` to version 1.17.0;
- `libx11` to version 1.8.10;
- `libxext` to version 1.3.6;
- `libice` to version 1.1.2;
- `libsm` to version 1.2.5;
- `libxt` to version 1.3.1;
- `libxmu` to version 1.2.1;
- `libxpm` to version 3.5.17;
- `libxaw` to version 1.0.16;
- `libxrender` to version 0.9.12;
- `libxcursor` to version 1.2.3;
- `libfontenc` to version 1.1.8;
- `libxfont2` to version 2.0.7;
- `libxi` to version 1.8.2;
- `libxtst` to version 1.2.5;
- `libxxf86vm` to version 1.1.6;
- `libxkbfile` to version 1.1.3;
- `libxshmfence` to version 1.3.3;
- `libxcvt` to version 0.1.3;
- `xkbcomp` to version 1.4.7;
- `xprop` to version 1.2.8;
- `appstream` to version 1.0.4.

This PR also updates the patch for `OpenSSL` to include risc-v support.

Supersedes #392 
Supersedes #390


docker/Dockerfile
```diff
@@ -11,6 +11,8 @@ RUN apt-get update \
 		autogen \
 		# Used by host-bison
 		autopoint \
+		# Used by the Linux kernel
+		bc \
```
avdgrinten, line 15: Why are we adding stuff used by the Linux kernel? We don't build the Linux kernel (?).

patches/appstream/0001-Add-Managarm-support.patch
```diff
@@ -1,42 +1,59 @@
...
+--- a/data/meson.build
++++ b/data/meson.build
+@@ -9,8 +9,8 @@ install_data('its/metainfo.loc',
+ 
+ # Do not rely on an exe wrapper for rel-info, use the system one in that case
+ if meson.is_cross_build()
+-    dependency('appstream', version: '>=' + as_version, native: true,
+-               not_found_message: 'Native appstream required for cross-building')
++#    dependency('appstream', version: '>=' + as_version, native: true,
++#               not_found_message: 'Native appstream required for cross-building')
```
avdgrinten, line 24: Why is this needed even if we have `host-appstream`?

patches/qt6/0001-Use-absolute-URLs-for-submodules.patch
```diff
@@ -1,401 +0,0 @@
...
---- a/.gitmodules
-+++ b/.gitmodules
-@@ -6,47 +6,47 @@
- [submodule "qtsvg"]
- 	depends = qtbase
- 	path = qtsvg
--	url = ../qtsvg.git
-+	url = https://code.qt.io/qt/qtsvg.git
- 	branch = 6.6.2
- 	status = addon
```
avdgrinten, line 22: Why is this patch dropped? The relative URLs are a problem on xbbs (where we pull from our mirror and not from upstream).

avdgrinten (approved):
LGTM except for some minor comments.

patches/qt6/0001-Use-absolute-URLs-for-submodules.patch
Dennisbonke: Because the QT superbuild system was changed and this change breaks the entire build. The superbuild will now deduce the base url part from the parent repo (the one we clone from the recipe) and will then add the url part to it, or so it seems. Surprisingly, this only happens for a submodule of qtdeclarative? I fought this for 2 hours and then gave up. The correct fix is to get us into upstream so we don't have to use patched repo's, the workaround in the meantime is to not pull this from the mirror (or mirror everything and hope for the best).

patches/appstream/0001-Add-Managarm-support.patch
Dennisbonke: Even with `host-appstream` meson refused to find the binary with `dependency`. As the program is in path, the `find_program` call works fine, finds host-appstream and the build will succeed after that.

docker/Dockerfile
Dennisbonke: That's right, Managarm does not do so. However, the `mlibc-lfs` project reuses the same rootfs, and they sure do build the Linux kernel. This is thus in preparation for the eventual move of `mlibc-lfs` to the Managarm org. It's also a small change which is why I included it.

PR #438, opened 2025-01-08
scripts/vm-util: add support for basic NVMe-over-fabrics

no92:
This assumes that the user spins up their own TFTP root and NVMe-over-fabrics controller, and thus could use more automation, but this is a simple start.


Dennisbonke (approved):
LGTM

avdgrinten:
We may want to create the TFTP directory for qemu in `vm-util.py`.

PR #439, opened 2025-01-20
vm-util: Disable UMIP if it is not supported on the host

Dennisbonke:
QEMU emulates UMIP regardless of whether or not it is supported. It does that by trapping every table access, including ltr, which we do on task switch when the I/O permission bitmap changes. This is quite costly in performance, so we disable UMIP if the host does not support it.


no92 (approved):
LGTM

PR #440, opened 2025-01-20
Fix minor bootstrap-managarm bugs

pitust:


bootstrap.d/app-emulation.y4.yml
```diff
@@ -24,6 +24,7 @@ tools:
         - '--enable-gtk'
         - '--enable-sdl'
         - '--enable-trace-backends=log'
+        - '--enable-slirp'
```
no92, line 27: This seems wrong - meson auto-detects libslirp without needing this flag. It only ever gets disabled when it's not present.

no92: Also, changing configure flags requires a revbump.

pitust: > This seems wrong - meson auto-detects libslirp without needing this flag. It only ever gets disabled when it's not present.
Weird, it didn't detect it for me (and this fixed it). libslirp is a subproject of qemu, though, so I suspect it's fetched automatically if you ask for it.

> Also, changing configure flags requires a revbump.
Done.

PR #444, opened 2025-02-05
systemd-udevd preparations

no92:


bootstrap.d/dev-python.y4.yml
```diff
@@ -0,0 +1,14 @@
...
+    labels: [aarch64, riscv64]
+    architecture: noarch
+    from_source: bragi
+    tools_required:
+      - host-python
+    revision: 1
+    install:
+      - args: ['python3', '-m', 'venv', '.']
+        workdir: '@PREFIX@'
+      - args: ['bin/pip', 'install', '-e', '@THIS_SOURCE_DIR@']
```
avdgrinten, line 12: `-e` doesn't work (the source dir will not be present when downloading the package from xbbs etc).

scripts/update-image.py
```diff
@@ -456,7 +456,7 @@ def generate_plan(arch, root_uuid, scriptdir):
     yield FsAction.RSYNC, "sbin"
     yield FsAction.RSYNC, "root/"
     yield FsAction.RSYNC, "usr/"
-    yield FsAction.RSYNC, "etc/"
+    yield FsAction.RSYNC, "etc/", ["udev/hwdb.bin"]
```
avdgrinten, line 459: Why do we need to preserve this?

no92: This would get regenerated on every boot by udev otherwise, and this takes a few seconds.

Dennisbonke (approved):
LGTM

PR #447, opened 2025-02-17
Systemd boot

Dennisbonke:
This PR adds the necessary patchwork to allow systemd to boot a Managarm system into Weston.
Additional patches enabling weston and kmscon will follow later.
For now, if one has it locally, one can specify `systemd` on the kernel commandline to force stage1 to exec into systemd, otherwise the stage2 fallback will be used.

Part of the systemd on Managarm project.

### DIFF
--- a/patches/systemd/0002-Add-support-for-systemd-boot-for-Managarm.patch
+++ b/patches/systemd/0002-Add-support-for-systemd-boot-for-Managarm.patch
@@ -1,0 +1,567 @@
+From c02304e2d1f9a5e95832104eb5c533afee403964 Mon Sep 17 00:00:00 2001
+From: Dennis Bonke <admin@dennisbonke.com>
+Date: Fri, 10 Jan 2025 00:55:19 +0100
+Subject: [PATCH 2/3] Add support for systemd boot for Managarm
+
+---
+ src/basic/capability-util.c              |  2 ++
+ src/basic/process-util.c                 |  8 ++++++--
+ src/basic/raw-clone.h                    |  6 ++----
+ src/basic/stdio-util.h                   |  1 +
+ src/core/exec-credential.c               |  4 ++++
+ src/core/exec-invoke.c                   | 11 ++++++++++
+ src/core/main.c                          |  3 +++
+ src/core/mount.c                         | 11 ++++++++++
+ src/journal/journald-native.c            |  3 +++
+ src/journal/journald-server.c            |  9 ++++++++
+ src/journal/journald-syslog.c            |  3 +++
+ src/journal/journald.conf                |  6 +++---
+ src/libsystemd/sd-journal/journal-send.c |  1 +
+ src/shared/async.c                       |  4 ++++
+ src/shared/mount-setup.c                 | 10 +++++++++
+ src/shared/mount-util.c                  |  7 +++++++
+ src/udev/udev.conf                       |  2 +-
+ units/systemd-journald.service.in        | 10 ++++-----
+ units/systemd-udevd.service.in           | 26 ++++++++++++------------
+ 19 files changed, 99 insertions(+), 28 deletions(-)
+
+diff --git a/src/basic/capability-util.c b/src/basic/capability-util.c
+index e9b41fe..ed8472c 100644
+--- a/src/basic/capability-util.c
++++ b/src/basic/capability-util.c
+@@ -333,6 +333,7 @@ int drop_privileges(uid_t uid, gid_t gid, uint64_t keep_capabilities) {
+         if (prctl(PR_SET_KEEPCAPS, 0) < 0)
+                 return log_error_errno(errno, "Failed to disable keep capabilities flag: %m");
+ 
++#ifndef __managarm__
+         /* Drop all caps from the bounding set (as well as the inheritable/permitted/effective sets), except
+          * the ones we want to keep */
+         r = capability_bounding_set_drop(keep_capabilities, true);
+@@ -365,6 +366,7 @@ int drop_privileges(uid_t uid, gid_t gid, uint64_t keep_capabilities) {
+                 if (cap_set_proc(d) < 0)
+                         return log_error_errno(errno, "Failed to increase capabilities: %m");
+         }
++#endif
+ 
+         return 0;
+ }
+diff --git a/src/basic/process-util.c b/src/basic/process-util.c
+index 6a1e496..48a5b76 100644
+--- a/src/basic/process-util.c
++++ b/src/basic/process-util.c
+@@ -1621,6 +1621,8 @@ int safe_fork_full(
+                 }
+         }
+ 
++// TODO: Managarm dies if the first or third argument to mount are null, untill we implement that, we should comment it out.
++#ifndef __managarm__
+         if (FLAGS_SET(flags, FORK_NEW_MOUNTNS | FORK_MOUNTNS_SLAVE)) {
+                 /* Optionally, make sure we never propagate mounts to the host. */
+                 if (mount(NULL, "/", NULL, MS_SLAVE | MS_REC, NULL) < 0) {
+@@ -1628,6 +1630,7 @@ int safe_fork_full(
+                         _exit(EXIT_FAILURE);
+                 }
+         }
++#endif
+ 
+         if (FLAGS_SET(flags, FORK_PRIVATE_TMP)) {
+                 assert(FLAGS_SET(flags, FORK_NEW_MOUNTNS));
+@@ -1973,8 +1976,9 @@ _noreturn_ void freeze(void) {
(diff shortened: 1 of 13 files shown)


patches/systemd-utils/0001-Add-managarm-support.patch
```diff
@@ -1,7 +1,7 @@
-From 7358a1c31a63ebdb2495cff5a26858319c27f116 Mon Sep 17 00:00:00 2001
```
no92, line 1: These patches are in the wrong subdirectory, `systemd-utils` instead of `systemd`.

patches/systemd/0002-Add-support-for-systemd-boot-for-Managarm.patch
```diff
@@ -0,0 +1,651 @@
...
+         if (FLAGS_SET(flags, FORK_PRIVATE_TMP)) {
+                 assert(FLAGS_SET(flags, FORK_NEW_MOUNTNS));
+@@ -1973,8 +1976,8 @@ _noreturn_ void freeze(void) {
+         }
+ 
+         /* waitid() failed with an unexpected error, things are really borked. Freeze now! */
+-        for (;;)
+-                pause();
++        for (;;);
++                // pause();
```
avdgrinten, line 81: Why is this patched out? This is on an error path anyway, isn't it?

patches/systemd/0002-Add-support-for-systemd-boot-for-Managarm.patch
```diff
@@ -0,0 +1,651 @@
...
+index 5201d5a..a4e5680 100644
+--- a/src/basic/time-util.c
++++ b/src/basic/time-util.c
+@@ -1704,7 +1704,7 @@ int time_change_fd(void) {
+ 
+         /* We only care for the cancellation event, hence we set the timeout to the latest possible value. */
+         static const struct itimerspec its = {
+-                .it_value.tv_sec = TIME_T_MAX,
++                .it_value.tv_sec = INT32_MAX, // HACK!
+         };
```
avdgrinten, line 125: Why is this needed? If this causes an overflow in mlibc or managarm-system, we should fix the overflow.

patches/systemd/0002-Add-support-for-systemd-boot-for-Managarm.patch
```diff
@@ -0,0 +1,651 @@
...
++++ b/src/core/exec-invoke.c
+@@ -226,8 +226,9 @@ static int connect_logger_as(
+         if (r < 0)
+                 return r;
+ 
+-        if (shutdown(fd, SHUT_RD) < 0)
+-                return -errno;
++        // TODO: Managarm does not implement shutdown
++        // if (shutdown(fd, SHUT_RD) < 0)
++        //         return -errno;
```
avdgrinten, line 172: This is potentially problematic (may break the other side of the pipe/socket/etc). We should at least log about this.

patches/systemd/0002-Add-support-for-systemd-boot-for-Managarm.patch
```diff
@@ -0,0 +1,651 @@
...
+diff --git a/src/core/manager.c b/src/core/manager.c
+index 22b1822..3b74cb0 100644
+--- a/src/core/manager.c
++++ b/src/core/manager.c
+@@ -926,6 +926,7 @@ int manager_new(RuntimeScope runtime_scope, ManagerTestRunFlags test_run_flags,
+                 .executor_fd = -EBADF,
+         };
+ 
++
+         unit_defaults_init(&m->defaults, runtime_scope);
```
avdgrinten, line 248: Nitpick: unnecessary hunk.

patches/systemd/0002-Add-support-for-systemd-boot-for-Managarm.patch
```diff
@@ -0,0 +1,651 @@
...
+--- a/src/core/mount.c
++++ b/src/core/mount.c
+@@ -1553,6 +1553,11 @@ static void mount_sigchld_event(Unit *u, pid_t pid, int code, int status) {
+                         f == MOUNT_SUCCESS,
+                         code, status);
+ 
++#ifdef __managarm__
++        if(m->state == MOUNT_MOUNTING && f == MOUNT_SUCCESS)
++                m->state = MOUNT_MOUNTING_DONE;
++#endif
```
avdgrinten, line 262: Why is this done? We should add a comment.

patches/systemd/0002-Add-support-for-systemd-boot-for-Managarm.patch
```diff
@@ -0,0 +1,651 @@
...
+                 return 0;
+ 
+-        r = server_determine_path_usage(s, storage->path, &vfs_used, &vfs_avail);
+-        if (r < 0)
+-                return r;
++        // r = server_determine_path_usage(s, storage->path, &vfs_used, &vfs_avail);
++        // if (r < 0)
++        //         return r;
++        vfs_used = 0;
++        vfs_avail = 16384;
```
avdgrinten, line 316: Why is this done? This should also get a comment.

patches/systemd/0002-Add-support-for-systemd-boot-for-Managarm.patch
```diff
@@ -0,0 +1,651 @@
...
+                 return log_error_errno(n, "Failed to read listening file descriptors from environment: %m");
+@@ -2817,9 +2821,11 @@ int server_init(Server *s, const char *namespace) {
+         if (r < 0)
+                 return r;
+ 
++#ifndef __managarm__
+         r = server_open_hostname(s);
+         if (r < 0)
+                 return r;
++#endif
```
avdgrinten, line 344: This is this done? Please add a comment.

patches/systemd/0002-Add-support-for-systemd-boot-for-Managarm.patch
```diff
@@ -0,0 +1,651 @@
...
+index bd043c8..524c550 100644
+--- a/src/shared/async.c
++++ b/src/shared/async.c
+@@ -86,6 +86,9 @@ static int close_func(void *p) {
+ }
+ 
+ int asynchronous_close(int fd) {
++#ifdef __managarm__
++        safe_close(fd); /* local fallback */
++#else
```
avdgrinten, line 410: Should we return something here? The current patch makes it return `-EBADF`.

patches/systemd/0002-Add-support-for-systemd-boot-for-Managarm.patch
Dennisbonke: A PR for this is posted, but with that implemented I hang during boot. Probably an issue of the implementation, but we can fix that and remove this piece of the patch later.

patches/systemd/0002-Add-support-for-systemd-boot-for-Managarm.patch
Dennisbonke: I've added a comment.

patches/systemd/0002-Add-support-for-systemd-boot-for-Managarm.patch
Dennisbonke: I've added a comment.

patches/systemd/0002-Add-support-for-systemd-boot-for-Managarm.patch
Dennisbonke: This is no longer needed and has been dropped.

patches/systemd/0002-Add-support-for-systemd-boot-for-Managarm.patch
Dennisbonke: This has been addressed.

patches/systemd/0002-Add-support-for-systemd-boot-for-Managarm.patch
Dennisbonke: I've added a comment.

patches/systemd/0002-Add-support-for-systemd-boot-for-Managarm.patch
```diff
@@ -0,0 +1,651 @@
...
+diff --git a/units/runsvr-usbhid.service b/units/runsvr-usbhid.service
+new file mode 100644
+index 0000000..5a171af
+--- /dev/null
++++ b/units/runsvr-usbhid.service
+@@ -0,0 +1,26 @@
++[Unit]
++Description=Cringe
++#Documentation=man:weston(1) man:weston.ini(5)
++#Documentation=https://wayland.freedesktop.org/
```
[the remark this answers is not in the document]
Dennisbonke: Done

PR #450, opened 2025-02-17
icu: Work around build failure on RISC-V

avdgrinten:
This is a potential multi lib issue but we can work around it for now.


Dennisbonke (approved):
LGTM

PR #455, opened 2025-02-20
mlibc: pin commit

no92:


avdgrinten:
This should be updated to the current mlibc HEAD before merging (as xbbs will probably not be happy about version downgrades).

avdgrinten:
LGTM otherwise. Also note that the pinned commit _must_ be on the `master` branch, not on `abi-break` for this to work.

PR #459, opened 2025-03-09
Gnulib updates

Dennisbonke:
This PR updates sed and m4 to fix gnulib related breakages after strerror_r was changed to provide both signatures as needed.


no92 (approved):
LGTM

PR #479, opened 2025-03-17
More systemd work

Dennisbonke:
This PR updates `core-files` to add more default users and groups that systemd uses, it also updates the patch to make use of `/proc/self/mountinfo` and `/proc/self/fdinfo` as support for that  has landed in Managarm.


no92 (approved):
LGTM

PR #488, opened 2025-03-29
rust: update rustc to 1.85.1 + partial stack upgrade

no92:


avdgrinten (approved):
It's a bit ugly that the auto generated patch is committed to the repo. Also, why are there so many changes in the autogenerated code?

Dennisbonke:
I concur with the autogenerated patch comment, can't we autogenerate that at runtime?

PR #501, opened 2025-04-06
OpenVi: Add port

johnsonjh:
OpenVi: Add port

### DIFF
--- a/bootstrap.d/app-editors.y4.yml
+++ b/bootstrap.d/app-editors.y4.yml
@@ -65,6 +65,41 @@ packages:
       - args: ['cp', '@SOURCE_ROOT@/extrafiles/nanorc', '@THIS_COLLECT_DIR@/home/managarm/.nanorc']
       - args: ['cp', '@SOURCE_ROOT@/extrafiles/root-nanorc', '@THIS_COLLECT_DIR@/root/.nanorc']
 
+  - name: openvi
+    architecture: '@OPTION:arch@'
+    metadata:
+      summary: Portable OpenBSD vi for UNIX systems
+      description: OpenVi is an enhanced and portable implementation of the Berkeley vi / ex text editor
+      spdx: 'BSD-3-Clause'
+      website: 'https://github.com/johnsonjh/OpenVi'
+      maintainer: "Jeffrey H. Johnson <trnsz@pobox.com>"
+      categories: ['app-editors']
+    source:
+      subdir: 'ports'
+      git: 'https://github.com/johnsonjh/OpenVi.git'
+      tag: '7.6.31'
+      version: '7.6.31'
+    tools_required:
+      - system-gcc
+    pkgs_required:
+      - mlibc
+      - ncurses
+    revision: 1
+    configure:
+      - args: ['cp', '-r', '@THIS_SOURCE_DIR@/.', '@THIS_BUILD_DIR@']
+    build:
+      - args: ['make', '-j@PARALLELISM@']
+        environ:
+          CC: "@OPTION:arch-triple@-gcc"
+          CFLAGS: "-DBYTE_ORDER=__BYTE_ORDER__"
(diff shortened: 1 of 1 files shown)


bootstrap.d/app-editors.y4.yml
```diff
@@ -65,6 +65,37 @@ packages:
...
+      summary: Portable OpenBSD vi for UNIX systems
+      description: OpenVi is an enhanced and portable implementation of the Berkeley vi / ex text editor
+      spdx: 'BSD-3-Clause'
+      website: 'https://github.com/johnsonjh/OpenVi'
+      maintainer: "Jeffrey H. Johnson <trnsz@pobox.com>"
+      categories: ['app-editors']
+    source:
+      subdir: 'ports'
+      git: 'https://github.com/johnsonjh/OpenVi.git'
+      tag: '7.6.31'
```
Dennisbonke, line 80: You're missing a version tag here.

bootstrap.d/app-editors.y4.yml
```diff
@@ -65,6 +65,37 @@ packages:
...
+    tools_required:
+      - system-gcc
+    pkgs_required:
+      - mlibc
+      - ncurses
+    build:
+      - args: ['cp', '-r', '@THIS_SOURCE_DIR@/.', '@THIS_BUILD_DIR@']
```
Dennisbonke, line 87: We normally do this in the configure step.

bootstrap.d/app-editors.y4.yml
```diff
@@ -65,6 +65,37 @@ packages:
...
+    source:
+      subdir: 'ports'
+      git: 'https://github.com/johnsonjh/OpenVi.git'
+      tag: '7.6.31'
+    tools_required:
+      - system-gcc
+    pkgs_required:
+      - mlibc
+      - ncurses
+    build:
```
Dennisbonke, line 86: Missing revision tag, it should be set to 1 for new ports. See the other ports in the file for how it should look. The vim port is a good candidate to look at.

Dennisbonke (changes requested):
Thanks for the PR! Looks good in general, a few comments tho.

johnsonjh:
I'll made the requested changes.

Just to mention also that by default this installs as `ovi`, `oex`, `oview`, etc.  If you set `BINPREFIX` to an empty string, the `o` prefix is dropped (in case you want to have it as the default `vi`).

johnsonjh:
I'm not sure if I need `labels` or not.  It should work everywhere without problems, but I only tested x86_64.

johnsonjh:
I notice the default temporary filename is /tmp/vi.000000 which is odd because it uses mkstemp.  But this would probably be an mlibc issue and not the port.

bootstrap.d/app-editors.y4.yml
```diff
@@ -65,6 +65,44 @@ packages:
...
+    configure:
+      - args: ['cp', '-r', '@THIS_SOURCE_DIR@/.', '@THIS_BUILD_DIR@']
+    build:
+      - args: ['make', '-j@PARALLELISM@']
+        environ:
+          CC: "@OPTION:arch-triple@-gcc"
+          CFLAGS: "-DBYTE_ORDER=__BYTE_ORDER__"
+          LDFLAGS: "-lncurses -ltinfo"
+          LTO: "1"
+      - args: ['make', 'strip']
```
no92, line 97: Not sure if we want to strip binaries?

johnsonjh: OK, I'll remove.

no92:
You might also want to squash the commits into one and use the `<package>: Add port` format for the commit message.

johnsonjh:
Squashed.  Not sure if I need / want labels.  I haven't explicitly tested the other platforms yet.

Dennisbonke (approved):
LGTM

PR #509, opened 2025-04-23
sqlite: Add missing soname

48cf:


bootstrap.d/dev-db.y4.yml
```diff
@@ -37,3 +37,5 @@ packages:
       - args: ['make', 'install']
         environ:
           DESTDIR: '@THIS_COLLECT_DIR@'
+      # Add an soname to the shared library
+      - args: ['patchelf', '--set-soname', 'libsqlte3.so.3.49.1', '@THIS_COLLECT_DIR@/usr/lib/libsqlite3.so.3.49.1']
```
ArsenArsen, line 41: does `--soname=legacy` not have the desirable effect?

48cf: You're right, fixed that

Dennisbonke (changes requested):
This should bump the revision as well.

Dennisbonke (approved):
LGTM

PR #523, opened 2025-07-03
scripts: update limine config

ElectrodeYT:
Update the Physical Hardware entries in the limine.conf file to log to serial using the userspace serial logging infrastructure, along with logging thor early boot logs to COM1.
Also use the systemd.target boot argument.

### DIFF
--- a/scripts/limine.conf
+++ b/scripts/limine.conf
@@ -50,12 +50,12 @@ timeout: 3
 		///managarm (Weston, serial output, plainfb)
 			kernel_path: boot():/managarm/eir-limine
 			protocol: limine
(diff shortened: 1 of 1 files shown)


scripts/limine.conf
```diff
@@ -50,12 +50,12 @@ timeout: 3
 		///managarm (Weston, serial output, plainfb)
 			kernel_path: boot():/managarm/eir-limine
 			protocol: limine
-			cmdline: serial init.launch=weston plainfb.force=1
+			cmdline: serial serial.dump=all init.launch=weston systemd.target=weston.target plainfb.force=1
```
no92, line 53: We do not use `weston.target`, this can be dropped.

PR #534, opened 2025-07-23
bootstrap: update URLs to match xbbsv2

no92:


no92:
I confirmed this as working for both tools and packages on the affected architectures.

PR #541, opened 2025-07-30
boot-artifacts: Add package

avdgrinten:
Add package with scripts to generate boot artifacts (e.g., the tftp directory for netboot) for various SoCs (though only the Raspberry Pi 4 is upstream right now).


Dennisbonke (commented):
In general looks good to me, but we should strive to keep the file sorted alphabetically (I know it's not the case everywhere but it's pretty decent). Can you move this to its proper location?

PR #547, opened 2025-08-05
Add alsa-utils, pipewire and wireplumber ports

Qwinci:
- Add ports for alsa-utils (depends on https://github.com/managarm/mlibc/pull/1399) + pipewire  + wireplumber + libpulse
- Add a pc file for lua

### DIFF
--- a/bootstrap.d/media-libs.y4.yml
+++ b/bootstrap.d/media-libs.y4.yml
@@ -40,6 +40,105 @@ packages:
         environ:
           DESTDIR: '@THIS_COLLECT_DIR@'
 
+  - name: libpulse
+    architecture: '@OPTION:arch@'
+    metadata:
+      summary: Libraries for PulseAudio clients
+      description: Libraries for PulseAudio clients.
+      spdx: 'LGPL-2.1-or-later'
+      website: 'https://www.freedesktop.org/wiki/Software/PulseAudio'
+      maintainer: "Qwinci <qwinci222@gmail.com>"
+      categories: ['media-libs']
+    source:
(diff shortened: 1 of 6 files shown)


bootstrap.d/media-libs.y4.yml
```diff
@@ -40,6 +40,101 @@ packages:
         environ:
           DESTDIR: '@THIS_COLLECT_DIR@'
 
+  - name: libpulse
+    architecture: '@OPTION:arch@'
+    metadata:
+      summary: Libraries for PulseAudio clients
+      description: Libraries for PulseAudio clients.
+      spdx: 'LGPL-2.1+'
```
Dennisbonke: The correct identifier is `LGPL-2.1-or-later`.

bootstrap.d/media-sound.y4.yml
```diff
@@ -0,0 +1,55 @@
...
+import:
+  - !managarm
+---
+packages:
+  - name: alsa-utils
+    architecture: '@OPTION:arch@'
+    metadata:
+      summary: Advanced Linux Sound Architecture Utils
+      description: This package contains the ALSA utilities.
+      spdx: 'GPL-2-only'
```
Dennisbonke: The correct identifier is `GPL-2.0-only`.

Dennisbonke (changes requested):
Some small items, and you're missing split-debug and the dbg subpackages for the new ports. Adding those in would be good. I see some other small style items and configure options that I want to check out but nothing that's blocking this from merging.

PR #604, opened 2025-11-10
libpulse: Add port

Qwinci:


bootstrap.d/media-libs.y4.yml
```diff
@@ -40,6 +40,101 @@ packages:
...
           DESTDIR: '@THIS_COLLECT_DIR@'
 
+  - name: libpulse
+    architecture: '@OPTION:arch@'
+    metadata:
+      summary: Libraries for PulseAudio clients
+      description: Libraries for PulseAudio clients.
+      spdx: 'LGPL-2.1+'
+      website: 'https://www.freedesktop.org/wiki/Software/PulseAudio'
+      maintainer: "Dennis Bonke <dennis@managarm.org>"
```
avdgrinten: Should this be Dennis? I think you can also just add yourself, Qwinci.

avdgrinten (approved):
LGTM but Dennis should probably also take a quick look.

bootstrap.d/media-libs.y4.yml
Dennisbonke: Yeah qwinci can add himself here but if he wants to pass maintainership to me that's fine too.

PR #679, opened 2026-04-05
libintl: avoid namespace pollution in `libintl.h` in strict POSIX mode

no92:
All packages depending on `libintl` continue to compile with this patch applied.


Dennisbonke (approved):
LGTM

PR #683, opened 2026-04-21
core-files: add basic `config.site`

no92:
This should be used to tell configure scripts about what stuff is _not_ broken on managarm, as often it is wrongly assumed that all unknown implementations are broken in specific ways.


Dennisbonke (approved):
LGTM

PR #693, opened 2026-05-18
KDE Plasma 6 part 9, the finale

Dennisbonke:
This PR adds SDDM and does various fixes to components used all around in KDE. We include SDDM by default and autoboot into Weston using SDDM as KDE isn't stable enough to make default, but the patchset is large enough that I don't want to keep it around myself. One can edit the sddm.conf file to disable autologin and get a login prompt, where one can then select KDE.

Known issues:
- No cursor in SDDM, which is why we autoboot.
- Logging in into KDE as !root is broken. Seems to be related to the dbus session bus?
- Logging in into Weston as !root works, but functionality is degraded. Seems to be related to the dbus session bus?
- On first boot into KDE, the panel doesn't show. A terminal can be launched with ctrl-shift-t assuming `konsole` is installed. On subsequent boots without regenerating the image the panel shows.
- The stability of KDE especially when running heavy applications like WebKitGTK has some improvements to make.
- Incidentially, during KDE startup, kwin_wayland dies, likely in QML JIT code. A reboot fixes it.


Mathewnd (approved):

qookei (approved):
I feel like this is out of my area of expertise, but I didn't see anything obviously wrong during a quick look through.

patches/kcmutils/0001-Disable-kirigami-check-for-tool-building.patch
```diff
@@ -0,0 +1,25 @@
...
+
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 02a87d9..acd68fc 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -71,7 +71,7 @@ ecm_set_disabled_deprecation_versions(
+     KF 6.23.0
+ )
+ 
+-ecm_find_qmlmodule(org.kde.kirigami REQUIRED)
```
no92, line 18: Why is this needed?

bootstrap.d/dev-libs.y4.yml
```diff
@@ -91,6 +79,26 @@ sources:
           '@BUILD_ROOT@/tools/host-automake-v1.15/share/automake-1.15/config.guess',
           '@THIS_SOURCE_DIR@/build-aux/']
 
+  - name: nss
```
no92, line 82: This commit is broken; why is this moving `nss`, adding `plasma-wayland-protocols` and changing the `wayland` source back to tarballs when we use git sources now?

no92 (commented):
Reviewed the first 8 commits up to and including `kdecoration`.

For reviewability it would be good to break this PR up into multiple pieces.

bootstrap.d/dev-libs.y4.yml
Dennisbonke: I blame rebasing due to your earlier Wayland upgrade that conflicted with mine. The end result looks sane to me, but I'm not surprised that this commit is jank then.

patches/kcmutils/0001-Disable-kirigami-check-for-tool-building.patch
Dennisbonke: Kirigami wasn't found even when it is installed. And it doesn't fail the build, the check is only there to ensure that kirigami, which is required at runtime, is present. We ensure its present at runtime due to the fact that the package depends on it, so I decided to remove the check here to avoid hours of cmake hell.

Dennisbonke:
After 8 parts of this, this PR should now be good to go and finalize KDE and SDDM into Managarm.

bootstrap.d/dev-libs.y4.yml
```diff
@@ -3049,11 +3050,11 @@ packages:
           DESTDIR: '@THIS_COLLECT_DIR@'
         quiet: true
       - args: ['mkdir', '-pv', '@THIS_COLLECT_DIR@/usr/lib/systemd/system/sockets.target.wants']
-      - args: ['mkdir', '-pv', '@THIS_COLLECT_DIR@/usr/lib/systemd/system/graphical.target.wants']
+      # - args: ['mkdir', '-pv', '@THIS_COLLECT_DIR@/usr/lib/systemd/system/graphical.target.wants']
       - args: ['cp', '-v', '@THIS_SOURCE_DIR@/weston.socket', '@THIS_COLLECT_DIR@/usr/lib/systemd/system']
       - args: ['cp', '-v', '@THIS_SOURCE_DIR@/weston.service', '@THIS_COLLECT_DIR@/usr/lib/systemd/system']
       - args: ['ln', '-svn', '../weston.socket', '@THIS_COLLECT_DIR@/usr/lib/systemd/system/sockets.target.wants/']
-      - args: ['ln', '-svn', '../weston.service', '@THIS_COLLECT_DIR@/usr/lib/systemd/system/graphical.target.wants/']
+      # - args: ['ln', '-svn', '../weston.service', '@THIS_COLLECT_DIR@/usr/lib/systemd/system/graphical.target.wants/']
```
avdgrinten: Delete these lines instead of commenting out.

patches/sddm/0003-HACK-Ignore-the-timestamp-of-the-config-file.patch
```diff
@@ -0,0 +1,30 @@
...
+@@ -177,9 +177,12 @@ namespace SDDM {
+ 
+         files << m_path;
+ 
++// HACK: We apparently don't do this correctly, so nuke it.
++#ifndef __managarm__
+         if (latestModificationTime <= m_fileModificationTime) {
+             return;
+         }
++#endif
```
avdgrinten, line 24: Please on an issue for this.

avdgrinten (approved):

PR #699, opened 2026-05-19
KDE Plasma 6 part 6

Dennisbonke:
As requested, splitting up https://github.com/managarm/bootstrap-managarm/pull/693 into chunks. This is part 6, adding several dependencies, nothing fancy here unfortunately.

Part of the KDE on Managarm project

Partly supersedes https://github.com/managarm/bootstrap-managarm/pull/693

### DIFF
--- a/bootstrap.d/dev-libs.y4.yml
+++ b/bootstrap.d/dev-libs.y4.yml
@@ -858,6 +858,62 @@ packages:
         environ:
           DESTDIR: '@THIS_COLLECT_DIR@'
 
+  - name: kirigami-addons
+    architecture: '@OPTION:arch@'
+    metadata:
+      summary: Visual end user components for Kirigami-based applications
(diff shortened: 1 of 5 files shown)


patches/qrencode/0001-Add-missing-strings.h-include.patch
```diff
@@ -0,0 +1,24 @@
...
+
+diff --git a/qrenc.c b/qrenc.c
+index c09c4ab..be89634 100644
+--- a/qrenc.c
++++ b/qrenc.c
+@@ -25,6 +25,7 @@
+ #include <stdio.h>
+ #include <stdlib.h>
+ #include <string.h>
++#include <strings.h>
```
no92: This has been made unnecessary in the meantime, we can drop this patch entirely.

no92 (commented):
Other than dropping one patch this is good.

PR #712, opened 2026-05-28
uhda: Add package

Qwinci:
Package uHDA in a package instead of a subproject.

### DIFF
--- a/bootstrap.d/managarm-system.y4.yml
+++ b/bootstrap.d/managarm-system.y4.yml
@@ -20,7 +20,8 @@ sources:
     subdir: 'ports'
     git: 'https://github.com/uDrivers/uHDA.git'
     branch: 'main'
-    commit: a631f3e7c794a7c8b55b3d54faa6636c224e26e6
+    rolling_version: true
+    commit: 56fa19579af0d57354045d05a609f51b3e881b0b
     version: '0.0pl@ROLLING_ID@'
 
   - name: managarm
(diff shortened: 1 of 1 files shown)


avdgrinten (approved):

bootstrap.d/managarm-system.y4.yml
```diff
@@ -250,6 +249,47 @@ packages:
...
+    labels: [aarch64, riscv64]
+    architecture: '@OPTION:arch@'
+    from_source: uhda
+    tools_required:
+      - host-llvm-toolchain
+      - host-mold
+      - system-gcc
+    pkgs_required:
+      - mlibc
+    configure:
```
Dennisbonke, line 270: Missing revision field.

bootstrap.d/managarm-system.y4.yml
```diff
@@ -250,6 +249,47 @@ packages:
...
+      description: This package provides a HDA driver.
+      spdx: 'BSD-3-Clause'
+      website: 'https://github.com/uDrivers/uHDA'
+      maintainer: "Qwinci <qwinci222@gmail.com>"
+      categories: ['sys-kernel']
+    labels: [aarch64, riscv64]
+    architecture: '@OPTION:arch@'
+    from_source: uhda
+    tools_required:
+      - host-llvm-toolchain
```
Dennisbonke, line 264: If we depend on llvm, why do we need system-gcc?

Dennisbonke (commented):
2 small items. For the rest it looks good to me.

bootstrap.d/managarm-system.y4.yml
Qwinci: Without it meson complains about missing binutils declared in the cross file, I am not completely sure whether its required otherwise but that's also something that should probably be changed for managarm-system if its going to be changed.

Dennisbonke: Hmm. Does adding cross-binutils instead of system-gcc solve the issue?

Qwinci: Looks like it does, I changed it to be like that.

Dennisbonke (approved):
LGTM

PR #722, opened 2026-06-01
Add a port of epiphany

Dennisbonke:


bootstrap.yml
```diff
@@ -307,3 +307,5 @@ container:
     - libxcb-util0-dev
     - libxcb-keysyms1-dev
     - libxcb-icccm4-dev
+    # Used by epiphany
+    - blueprint-compiler
```
Qwinci, line 311: Looks like Epiphany also needs `desktop-file-utils` (for `update-desktop-database`) during build.

Qwinci (changes requested):
Looks good to me apart from a missing desktop-file-utils needed during build.

bootstrap.yml
Dennisbonke: Ah yes. That's me missing a patch as that's completely useless during cross compilation. It also reminds me to add it as a post-install step. Good catch. I'll fix it.

PR #725, opened 2026-06-10
sdl3: Add port

xinitrcn1:
some apps would like it

### DIFF
--- a/bootstrap.d/media-libs.y4.yml
+++ b/bootstrap.d/media-libs.y4.yml
@@ -2119,6 +2119,87 @@ packages:
           DESTDIR: '@THIS_COLLECT_DIR@'
         quiet: true
 
+  - name: sdl3
+    architecture: '@OPTION:arch@'
+    metadata:
+      summary: Simple Direct Media Layer version 3
+      description: The Simple DirectMedia Layer Version 3 (SDL3 for short) is a cross-platform library designed to make it easy to write multimedia software, such as games and emulators.
+      spdx: 'Zlib'
+      website: 'https://www.libsdl.org/'
+      maintainer: "Dennis Bonke <dennis@managarm.org>"
+      categories: ['media-libs']
+    source:
+      subdir: 'ports'
+      git: 'https://github.com/libsdl-org/SDL.git'
+      tag: 'release-3.4.10'
+      version: '3.4.10'
+    tools_required:
+      - system-gcc
+      - host-cmake
+      - host-pkg-config
+      - wayland-scanner
+    pkgs_required:
+      - mlibc
+      - libdrm
+      - libiconv
+      - libx11
+      - libxcursor
+      - libxext
+      - libxfixes
+      - libxi
+      - libxinerama
+      - libxkbcommon
+      - libxrandr
+      - libxrender
+      - libxscrnsaver
+      - libxxf86vm
+      - mesa
+      - systemd
+      - wayland
+      - wayland-protocols
+    revision: 1
+    configure:
+      - args:
+        - 'cmake'
+        - '-GNinja'
+        - '-DCMAKE_TOOLCHAIN_FILE=@SOURCE_ROOT@/scripts/CMakeToolchain-@OPTION:arch-triple@.txt'
+        - '-DCMAKE_INSTALL_PREFIX=/usr'
+        - '-DSDL_ALTIVEC=OFF'
+        - '-DSDL_DISKAUDIO=OFF'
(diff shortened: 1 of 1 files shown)


marv7000:
Any concrete examples of dependents?

xinitrcn1:
> Any concrete examples of dependents?

myself (Eden).

Ok maybe that doesn't count. How about:

- shadPS4 [not here yet]
- vita3k [not here yet]
- luanti [not here yet]
- raylib [optional, also not here yet]
- FreeRDP [not here]
- pcsx2 [not here]

Fair most of them aren't even in here, but it should make the life easier of whoever is porting them. Especially since SDL3 and SDL2 pretty much just "work" similarly - bar some differences with udev/input/etc. Probably most interesting at this immediate moment would be FreeRDP I'd assume?

bootstrap.d/media-libs.y4.yml
```diff
@@ -2119,6 +2119,88 @@ packages:
           DESTDIR: '@THIS_COLLECT_DIR@'
         quiet: true
 
+  - name: sdl3
+    labels: [aarch64, riscv64]
```
no92: Was compiling this tested on aarch64 and riscv64?

bootstrap.d/media-libs.y4.yml
```diff
@@ -2119,6 +2119,88 @@ packages:
...
+      - libxkbcommon
+      - libxrandr
+      - libxrender
+      - libxxf86vm
+      - mesa
+      - systemd
+      - wayland
+      - wayland-protocols
+      - libxscrnsaver
+    revision: 4
```
no92: Revision should be 1.

bootstrap.d/media-libs.y4.yml
```diff
@@ -2119,6 +2119,88 @@ packages:
...
+        - 'cmake'
+        - '-GNinja'
+        - '-DCMAKE_TOOLCHAIN_FILE=@SOURCE_ROOT@/scripts/CMakeToolchain-@OPTION:arch-triple@.txt'
+        - '-DCMAKE_INSTALL_PREFIX=/usr'
+        - '-DSDL_ALTIVEC=OFF'
+        - '-DSDL_DISKAUDIO=OFF'
+        - '-DSDL_OPENGL=ON'
+        - '-DSDL_OPENGLES=ON'
+        - '-DSDL_PTHREADS=ON'
+        - '-DSDL_PTHREADS_SEM=OFF'
```
no92: Is this actually used? What sort of issue does this run into?

From looking at the source, the only explanation I saw was that we stub `sem_getvalue`, which we should just fix in mlibc.

Dennisbonke:
Ah, that's where I recognized the GitHub handle from. That solves the question who will use it. I'll give a more in-depth review later. Quick note, while in this case it's fine to list me as the maintainer, usually something like that should be discussed first :^)

bootstrap.d/media-libs.y4.yml
```diff
@@ -2119,6 +2119,88 @@ packages:
...
+      tag: 'release-3.4.10'
+      version: '3.4.10'
+    tools_required:
+      - system-gcc
+      - host-cmake
+      - host-pkg-config
+      - wayland-scanner
+    pkgs_required:
+      - mlibc
+      - libdrm
```
no92, line 2143: Style nit: we usually put mlibc at the top, and sort the rest lexicographically.

no92 (commented):
The two commits should be squashed into one, and the commit message should follow the generat tyle of `package: short description`, e.g. `sdl3: Add port`.

bootstrap.d/media-libs.y4.yml
no92: `sem_getvalue` should be implemented correctly by managarm/mlibc#1799.

bootstrap.d/media-libs.y4.yml
xinitrcn1: Not at all. Disabled for now.

bootstrap.d/media-libs.y4.yml
xinitrcn1: When I built it I had scary errors w.r.t to thread sem, will re-enable.

xinitrcn1:
> Ah, that's where I recognized the GitHub handle from. That solves the question who will use it. I'll give a more in-depth review later. Quick note, while in this case it's fine to list me as the maintainer, usually something like that should be discussed first :^)

Oh sorry about that. I'll keep that in mind next time.

PR #729, opened 2026-06-12
bragi-bin: install bragi binary on managarm

no92:


avdgrinten (approved):

Dennisbonke (changes requested):
NAK, we have python-bragi already and as far as I can see that is the _exact_ same.

PR #734, opened 2026-06-15
fmt: Add port

xinitrcn1:
Trivial port, you know who's the user of this.
This time added myself as maintainer yay! :D

Tested on x86_64 only.

### DIFF
--- a/bootstrap.d/dev-libs.y4.yml
+++ b/bootstrap.d/dev-libs.y4.yml
@@ -564,6 +564,46 @@ packages:
       - name: elfutils-dbg
         include: ['/usr/lib/debug/**']
 
+  - name: libfmt
+    architecture: '@OPTION:arch@'
+    metadata:
+      summary: C++ modern format library
+      description: A modern formatting library.
+      spdx: 'MIT'
+      website: 'https://fmt.dev'
+      maintainer: "Elizabeth P. <lizzie@eden-emu.dev>"
+      categories: ['dev-libs']
+    source:
+      subdir: 'ports'
+      url: 'https://github.com/fmtlib/fmt/archive/refs/tags/12.1.0.tar.gz'
+      format: 'tar.gz'
+      checksum: 'blake2b:9e9cc77bd6f5ec31f0b43091e54eb508349d278e7873bcdff4eb9388d78584467da5837bdd10de1de84feb8ab0d63880df1c9fe0f6f0db40e47598646179b06e'
+      extract_path: 'fmt-12.1.0'
+      version: '12.1.0'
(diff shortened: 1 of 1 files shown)


bootstrap.d/dev-libs.y4.yml
```diff
@@ -564,6 +564,42 @@ packages:
       - name: elfutils-dbg
         include: ['/usr/lib/debug/**']
 
+  - name: fmt
```
Dennisbonke: I'd name it `libfmt`, to match gentoo for example.

bootstrap.d/dev-libs.y4.yml
```diff
@@ -564,6 +564,42 @@ packages:
...
+    pkgs_required:
+      - mlibc
+    revision: 1
+    configure:
+      - args:
+        - 'cmake'
+        - '-GNinja'
+        - '-DCMAKE_TOOLCHAIN_FILE=@SOURCE_ROOT@/scripts/CMakeToolchain-@OPTION:arch-triple@.txt'
+        - '-DCMAKE_INSTALL_PREFIX=/usr'
+        - '-DCMAKE_BUILD_TYPE=Release'
```
Dennisbonke: We try to default to RelWithDebInfo for cmake projects, and the equivalent for meson and friends.

bootstrap.d/dev-libs.y4.yml
```diff
@@ -564,6 +564,42 @@ packages:
...
+        - '@THIS_SOURCE_DIR@'
+    build:
+      - args: ['ninja']
+      - args: ['ninja', 'install']
+        environ:
+          DESTDIR: '@THIS_COLLECT_DIR@'
```
Dennisbonke, line 601: Missing split-debug and the corresponding debug package.

Dennisbonke (changes requested):
Some comments

xinitrcn1:
Just for future reference.
The package names should match those found on Gentoo? And if it's a library in dev-libs it should have the aforementioned split debug thingies?

Dennisbonke:
It's not a hard rule but usually yes. We source our names from (B)LFS and gentoo usually, falling back to Debian if they don't package it and making it up ourselves otherwise. Regarding split-debug, _every_ package that installs more than just headers (which I believe fmt does?) should have a split-debug package, and building in RelWithDebInfo (or equivalent) should be done for all packages regardless (yes there are plenty of packages that don't do that yet, the conversion to split-debug is ongoing, and a few packages are exempt for various reasons, usually for producing 20 gigs of debug info).

PR #735, opened 2026-06-16
scripts/vm-util: add `qemu_virtio_device` helper for virtio devices

no92:
This also avoids errors on Python <= 3.12, as observed on xbbs.

### DIFF
--- a/scripts/vm-util.py
+++ b/scripts/vm-util.py
@@ -47,6 +47,18 @@ def qemu_check_device(qemu, args, dev, fatal=False):
         sys.exit(1)
     return False
 
+def qemu_virtio_device(device, extra_args, iommu, legacy=False, modern=True):
+    parts = [
+        f"disable-modern={'off' if modern else 'on'}",
+        f"disable-legacy={'off' if legacy else 'on'}",
+    ]
+    if modern and not legacy:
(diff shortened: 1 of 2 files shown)


scripts/vm-util.py
```diff
@@ -47,6 +47,17 @@ def qemu_check_device(qemu, args, dev, fatal=False):
         sys.exit(1)
     return False
 
+def qemu_virtio_device(device, extra_args, iommu, modern = True):
+    parts = []
+    if extra_args:
+        parts.extend(extra_args)
+    if modern:
+        parts.append("disable-legacy=on")
```
avdgrinten: Why are we passing `disable-legacy=on`? It probably makes more sense to use transitional devices (fewer qemu flags and more likely to match the situation that users will encounter when running without vm-util.py).

no92: Legacy transport devices are really undesirable, as they are limited in their capabilities, only work on x86, always silently bypass IOMMU, and are long-deprecated in favor of the virtio specification which introduced modern transport. Making legacy opt-in seems reasonable to me.

avdgrinten: Legacy devices won't be used anyway unless you pass `disable-modern=on`.

avdgrinten (approved):

PR #738, opened 2026-06-18
scripts: Custom `xbstrap` command prefix

viktorpopp:
So my clangd was failing because `xbstrap` wasn't found. Initially i just added `uvx` before running `xbstrap`, but i thought i would create this PR because it would be easier to add the fix to the [Troubleshooting the LSP setup](https://docs.managarm.org/handbook/contributing/lsp.html#troubleshooting) section of the handbook.

Also someone should probably check that this works completely, and that my device doesn't do some quirky thing that makes it work, even tho it should work on all devices (i think).

### DIFF
--- a/scripts/managarm-lsp-launcher.sh
+++ b/scripts/managarm-lsp-launcher.sh
@@ -1,7 +1,28 @@
 #!/bin/sh
+
+# parse options
+XBSTRAP="xbstrap"
+BUILD_DIR=""
+while [ $# -gt 0 ]; do
+    case "$1" in
+        --xbstrap-bin=*)
+            XBSTRAP="${1#*=}"
+            shift 1
+            ;;
+        *)
+            if [ -z "$BUILD_DIR" ]; then
+                BUILD_DIR="$1"
+            fi
+            shift 1
+            ;;
+    esac
+done
+BUILD_DIR="${BUILD_DIR:-/dev/null}"
+
 set -ue
-# check if $1 is a valid build directory
-[ -L "${1:-/dev/null}"/bootstrap.link ] || exit 1
+
+# check if the build directory is valid
+[ -L "$BUILD_DIR"/bootstrap.link ] || exit 1
 
 # find compile_commands.json
 sourcedir="$(pwd)"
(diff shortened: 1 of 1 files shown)


no92:
In my VSCodium, I can use with with the following config:

```
		"clangd.path": "/home/leo/Documents/managarm/src/scripts/managarm-lsp-launcher.sh",
		"clangd.arguments": ["/home/leo/Documents/managarm/build/", "uvx --with y4"],
```

ArsenArsen:
it may be more reasonable to pass `XBSTRAP` via env, and allow that to be split by `IFS`.  so, `: "${XBSTRAP:=xbstrap}"` and `exec $XBSTRAP`.

this works for venvs (`XBSTRAP=/path/to/venv/bin/xbstrap launcher`) as well as for this case (`XBSTRAP='uvx xbstrap' launcher`), where this has no way to cover the former case

viktorpopp:
That was my initial idea, but i cannot find a way to pass environment variables the the lsp launcher script when using VSCode.

scripts/managarm-lsp-launcher.sh
```diff
@@ -1,7 +1,26 @@
...
+            XBSTRAP="${1#*=}"
+            shift 1
+            ;;
+        *)
+            OTHER_ARGS+=("$1")
+            shift 1
+            ;;
+    esac
+done
+BUILD_DIR=${OTHER_ARGS[0]:-/dev/null}
```
no92: This is currently marked as a `/bin/sh` script, so we should avoid bash-specific stuff:

```suggestion
# parse options
XBSTRAP="xbstrap"
BUILD_DIR=""
while [ $# -gt 0 ]; do
    case "$1" in
        --xbstrap-bin=*)
            XBSTRAP="${1#*=}"
            shift 1
            ;;
        *)
            if [ -z "$BUILD_DIR" ]; then
                BUILD_DIR="$1"
            fi
            shift 1
            ;;
    esac
done
BUILD_DIR="${BUILD_DIR:-/dev/null}"
```

no92 (approved):

PR #757, opened 2026-08-06
sys-apps: Use pkgconfig-for-target to util-linux

qookei:
This fixes util-linux not being able to find ncurses and disabling dmesg as a result.

Also disable lsblk (which wasn't previously built) because it does not compile due to missing BLKROGET ioctl.


avdgrinten (approved):

Dennisbonke (approved):
LGTM

